PR #1, opened 2019-07-09
Add detailed readme to users service

mukobi:
Added a README.md specifically for the users service which can be copied and adopted for other services.


mukobi:
> I'm not familiar with the details of licensing - is the MIT license the right one to use in this case?
> Everything else looks good

As I understand it, the MIT license is pretty simple and permissive, but maybe something else is better. Also, I put Google LLC in the license copyright, but maybe it should be something else instead? Adding Evan for review of this.

evankanderson:
I would recommend the Apache license, since that is what the rest of Knative is authored with.

See https://github.com/knative/serving/blob/master/LICENSE and https://github.com/knative/serving/tree/master/hack/boilerplate/ for the LICENSE file and a script to add the correct boilerplate to the top of each file.

mukobi:
@evankanderson Just switched to/added the Apache license.

PR #2, opened 2019-07-10
Users mongodb atlas

mukobi:
Connect users service to MongoDB Atlas, hook up querying a user's authorization level from the db, and integrate tests of the database.


mukobi:
The CI checks are failing because Travis is hiding the database password, an encrypted environment, from the test build on this PR since it's coming from my fork of this repo--see https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions. Is there a better way to handle this? The Travis build passes on my fork: https://travis-ci.com/mukobi/large-events

mukobi:
Testing/Travis-CI update: I went ahead and mocked away the db in testing using the mongomock Python module. This module closely emulates the behaviour of the pymongo module so the tests should remain safe and robust. I know mocking is not always ideal, but now the tests all pass (and run faster).

users/app.py
```diff
@@ -22,31 +22,27 @@
 """
 
 import os
-
 from flask import Flask, jsonify, request
+import pymongo
 
 app = Flask(__name__)  # pylint: disable=invalid-name
 
+DB = None  # db initialized in main runtime
```
evankanderson: Have you tested this with `gunicorn` or running in a container?

I think that `gunicorn` imports your module but it's no longer named `__main__`, so you might need to either use a function to initialize this globally, or add an `@app.before_first_request` method that will lazy-initialize on the first request.

(Or, I did more work than I needed last year when I was doing something like this with Cloud Functions).

users/app.py
```diff
@@ -60,5 +56,40 @@ def add_update_user():
...
+
+    Documents in the users collection should look like
+        {"username": "cmei4444",
+        "name": "Carolyn Mei",
+        "is_organizer": True}
+    """
+    if users_collection.count_documents({"username": username}) == 0:
+        return False  # user not found
+    cursor = users_collection.find({"username": username})
+    for user in cursor:
```
evankanderson: If you expect only a single document, you could also take the first item and then assert that the cursor was done.

This comes down to whether you expect to allow multiple documents with the same username, or if that indicates some sort of underlying logic error.

users/test.py
```diff
@@ -20,17 +20,43 @@
...
-        self.assertTrue(is_authorized_to_edit("carolyn", None))
-
-    def test_bad_user_not_authorized(self):
-        """An non-authorized user should not have authorized privileges"""
-        self.assertFalse(is_authorized_to_edit("Voldemort", None))
+    """Test /authorization endpoint of users service."""
+
+    def setUp(self):
+        """Seed mock DB for testing"""
+        self.id_of_test_user_is_authorized = "authorized-user"
```
evankanderson: For these sorts of constants, a global is usually fine, e.g. at the top level:

AUTHORIZED_USER = "authorized-user"
INTRUDER = "non-authorized-user"
MISSING = "not-a-user-in-the-database"

users/Dockerfile
```diff
@@ -22,7 +22,7 @@ WORKDIR $APP_HOME
 COPY . .
 
 # Install production dependencies.
-RUN pip install Flask gunicorn
+RUN pip install -r requirements.txt
```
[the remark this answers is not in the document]
mukobi: Got it, will do in the future.

users/app.py
mukobi: You're right, it doesn't initialize when deployed to Cloud Run. The reason I wasn't initializing this globally is because Travis-CI protects environmental secrets from pull requests from forks, so in order to get the tests the run I needed the app to not try to connect to the database.

I can think of 2 ways to solve this: @app.before_first_request as you suggested which probably works with few changes but looks a little fancy or refactoring the helper functions from app.py to some other imported module and only testing that module which maybe is better for organization but sounds like a pain and needless complication. Do you have a suggestion for which path to take?

users/app.py
mukobi: Ideally, for authorization purposes we should have the identifying user information be unique to each user (I think OAuth 2.0 auto generates unique user IDs). The MongoDB collection.find query returns a cursor with all results and not just the first, so I think I'll switch to checking just the first result if that's what you mean.

users/test.py
mukobi: Done

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/2/commits/d5c146c0e92b1c86c7f14a7413fcf05905089807 https://github.com/knative-portability/large-events/pull/2/commits/b9ae63e97ced0c21d15a2164be1b3a534aec6760 https://github.com/knative-portability/large-events/pull/2/commits/ddcdbff0d00629f99071e41e21379025b350ae91

users/app.py
```diff
@@ -60,5 +54,43 @@ def add_update_user():
...
+    """Queries the db to find authorization of the given user.
+
+    Documents in the users collection should look like
+        {"username": "cmei4444",
+        "name": "Carolyn Mei",
+        "is_organizer": True}
+    """
+    if users_collection.count_documents({"username": username}) == 0:
+        return False  # user not found
+    cursor = users_collection.find({"username": username})
```
evankanderson: Should this use [`find_one`](https://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.find_one)?

users/app.py
```diff
@@ -60,5 +54,43 @@ def add_update_user():
...
+    first_user = cursor[0]
+    authorized = first_user.get("is_organizer")
+    return False if authorized is None else authorized
```
evankanderson: I think you can compare with `True` or use the `bool` constructor to convert `None` to `False`:

```python
return bool(user.get("is_organizer"))
```

users/app.py
```diff
@@ -60,5 +54,43 @@ def add_update_user():
...
+
+
+@app.before_first_request
+def initialize_mongodb():
+    """Connect to MongoDB Atlas database using env vars.
```
evankanderson: Is this limited to Atlas (their cloud product), or simply using Atlas in your case? If it's the latter, I'd just say "Connect to MongoDB instance using env vars".

users/app.py
```diff
@@ -60,5 +54,43 @@ def add_update_user():
...
+
+    app.before_first_request decorator is used to keep
+    the DB from initializing in testing to remove the
+    dependency on the MongoDB connection in tests.
+    """
+    mongodb_atlas_username = os.environ.get(
```
evankanderson: Since these are local variables, and the name of the function is "...mongodb", you can probably use shorter names here, like `username`, `password`, and `cluster`.

users/app.py
```diff
@@ -60,5 +54,43 @@ def add_update_user():
...
+        "MONGODB_ATLAS_USERNAME")
+    mongodb_atlas_password = os.environ.get(
+        "MONGODB_ATLAS_PASSWORD")
+    mongodb_atlas_cluster_address = os.environ.get(
+        "MONGODB_ATLAS_CLUSTER_ADDRESS")
+    mongodb_uri = "mongodb+srv://{}:{}@{}".format(
```
evankanderson: Since you're using python3.7, consider using an f-string here:

`f"mongodb+srv://{username}:{password}@{cluster}"`

users/test.py
```diff
@@ -20,17 +20,51 @@
 """
 
 import unittest
-from app import is_authorized_to_edit
+import mongomock
+from app import find_authorization_in_db
+
+AUTHORIZED_USER = "authorized-user"
+NON_AUTHORIZED_USER = "non-authorized-user"
+MALFORMATTED_IN_DB_USER = "this-user-has-no-'is_organizer'_db_field"
```
evankanderson: Good test case!

users/app.py
mukobi: Fixed using @app.before_first_request that lazily inits a global DB object.

users/app.py
mukobi: Ah, thank you. Fixed https://github.com/knative-portability/large-events/pull/2/commits/00a50b4cf0f70b251b63ffe424875b6979401b2e

users/app.py
mukobi: Fixed https://github.com/knative-portability/large-events/pull/2/commits/a2adc51ddc490f7dbca20b6abe1833efe75bb784

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/2/commits/1676bea8fdd14a825b2efc42215a85e9c2b18571

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/2/commits/1676bea8fdd14a825b2efc42215a85e9c2b18571

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/2/commits/32034f2b9b35fc5b2d3d961401b7e4b84671b1a3

mukobi: Actually, just changed the whole function to be simpler https://github.com/knative-portability/large-events/pull/2/commits/b943855d9abb766fafbe7e91de3bd0e173ce491d

users/app.py
mukobi: Actually, just changed the whole function to be simpler https://github.com/knative-portability/large-events/pull/2/commits/b943855d9abb766fafbe7e91de3bd0e173ce491d

mukobi:
@evankanderson I resolved all your comments and linked the corresponding commits. Also tested in a container and using gunicorn on Cloud Run. Good to merge?

evankanderson:
Yes, good to merge.

users/app.py
```diff
@@ -60,5 +54,33 @@ def add_update_user():
...
+    """Queries the db to find authorization of the given user.
+
+    Documents in the users collection should look like
+        {"username": "cmei4444",
+        "name": "Carolyn Mei",
+        "is_organizer": True}
+    """
+    if users_collection.count_documents({"username": username}) == 0:
+        return False  # user not found
+    first_user = users_collection.find_one({"username": username})
```
evankanderson: It looks like `find_one` will return `None` if there are no matching documents, so you shouldn't need to call `count_documents` above. (This is more efficient on the database side, as it will do only one search.)

mukobi: Done https://github.com/knative-portability/large-events/pull/2/commits/249cd4512fc9efe0a420fbecd5200be21b73e141

PR #4, opened 2019-07-11
Created events skeleton and Event class

cmei4444:
Skeleton of all endpoints for the events service, connected to the DB.
Also added basic web pages for viewing events and posts with example data.


events/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# https://hub.docker.com/_/python
+FROM python:3.7
+
+# Copy local code to the container image.
+ENV APP_HOME /app
+WORKDIR $APP_HOME
+COPY . .
+
+# Install production dependencies.
+RUN pip install Flask gunicorn pymongo[srv]
```
evankanderson: Gabe just switched the other path to use a `requirements.txt` in #2 

You might want to do that as well.

events/app.py
```diff
@@ -0,0 +1,113 @@
...
+MONGODB_ATLAS_PASSWORD = os.environ.get(
+    "MONGODB_ATLAS_PASSWORD")
+MONGODB_ATLAS_CLUSTER_ADDRESS = os.environ.get(
+    "MONGODB_ATLAS_CLUSTER_ADDRESS")
+PYMONGO_URI = "mongodb+srv://{}:{}@{}".format(
+    MONGODB_ATLAS_USERNAME,
+    MONGODB_ATLAS_PASSWORD,
+    MONGODB_ATLAS_CLUSTER_ADDRESS)
+# print("Pymongo URI: {}".format(PYMONGO_URI))
+client = pymongo.MongoClient(PYMONGO_URI)
```
evankanderson: This is a different way of building the Mongo client compared with #2. You and Gabe should compare notes, and consider a standard way of handling this. (His has a `if __name__ == '__main__'` block that does the initialization.)

events/app.py
```diff
@@ -0,0 +1,113 @@
...
+    event.add_to_db()
+    return Response(
+        status=201,
+    )
+
+
+def build_event_info(info, time):
+    """Adds created_at time to event info dict"""
+    # MongoDB timestamp precision level, floor to milliseconds
+    time -= datetime.timedelta(0, 0, time.microsecond % 1000)
```
evankanderson: You can just assign to `time.microsecond`, FYI.

What does mongodb do if you pass a higher-precision timestamp? Does it error, or just throw away data? If the latter, you might not need to do this.

events/app.py
```diff
@@ -0,0 +1,113 @@
...
+testDB = client.test
+eventsDB = client.eventsDB
+events_coll = eventsDB.all_events
+
+
+@app.route('/v1/add', methods=['POST'])
+def add_event():
+    # TODO(cmei4444): Get event information from request data
+    info = request.form['info']
+    # Add information into database
```
evankanderson: These sorts of comments can be confusing, since the following 9 lines don't actually add the information to the database if I read it correctly. The actual add happens in `event.add_to_db`, I presume.

events/app.py
```diff
@@ -0,0 +1,113 @@
...
+    current_time = datetime.datetime.now()
+    build_event_info(info, current_time)
+    event = Event(info)
```
evankanderson: Consider having the Event constructor default or overwrite the 'created_at' parameter if not present in `info`.

events/app.py
```diff
@@ -0,0 +1,113 @@
...
+        'name',
+        'description',
+        'author',
+        'created_at',
+        'event_time']
+
+    def __init__(self, info):
+        self.info = dict(info)
+        # event_id is defined by the database, not the user
+        if "event_id" not in self.info:
```
evankanderson: Do you want to allow `event_id` to be set by the caller?

events/app.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+@app.route('/v1/<event_id>', methods=['PUT'])
+def get_one_event(event_id):
+    """Retrieve one event by event_id"""
+    pass
+
+
+class Event(object):
+    attributes = [
```
evankanderson: What happens if someone inserts an `info` which has fields other than the ones in `attributes`?

events/requirements.txt
```diff
@@ -0,0 +1,2 @@
+flask
+pymongo[srv]
```
evankanderson: This is different than the Dockerfile line 11?

events/test.py
```diff
@@ -0,0 +1,92 @@
+import unittest
+import datetime
+import app
+
+
+class TestEventsDB(unittest.TestCase):
```
evankanderson: You might want to use the `mongomock` (which is actually a fake) that Gabe found in #2.

evankanderson: Ref for differences between Mocks, Fakes, and Stubs:

https://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs

events/test.py
```diff
@@ -0,0 +1,92 @@
...
+        self.test_event = app.Event(self.event_info)
+        # make sure DB has no other test entries
+        app.events_coll.delete_many(self.event_info)
+
+    def test_add(self):
+        self.test_event.add_to_db()
+        queried_event = app.events_coll.find_one(self.event_info)
+        self.assertEqual(app.Event(queried_event), self.test_event)
+
+    def test_build_info(self):
```
evankanderson: It's best for each test to test exactly one thing. In this case, it seems like this test is testing the time rounding behavior, so I might call it `test_time_rounding'.

events/test.py
```diff
@@ -0,0 +1,92 @@
...
+        self.assertEqual(info['created_at'], time)
+
+    def tearDown(self):
+        app.events_coll.delete_many(
+            self.event_info)    # empty DB of test entries
+
+
+class TestEventsClass(unittest.TestCase):
+    def setUp(self):
+        # Sets up two events with identical data, one with event_id and one
```
evankanderson: Again, these notes can be great while you're figuring out what you want to write, but you usually want to remove them once the code is written.

People will read the comments and believe the code below does what is described, even as the code below changes. So it's best to use comments to describe **why** a certain pattern is needed, or other places where the code is subtle.

events/test.py
```diff
@@ -0,0 +1,92 @@
...
+    def test_construct_event(self):
+        event = app.Event(self.test_info)
+        self.test_info['event_id'] = None
+        self.assertEqual(event.info, self.test_info)
+
+        event_with_id = app.Event(self.test_info_with_id)
+        self.assertEqual(event_with_id.info, self.test_info_with_id)
+
+    def test_events_equal(self):
+        test_info_diff = {'name': 'different event',
```
evankanderson: This might be clearer as:

```python
test_info_diff = self.test_info.copy()
test_info_diff['name'] = 'different event'
```

events/test.py
```diff
@@ -0,0 +1,92 @@
...
+        event_diff = app.Event(test_info_diff)
+        self.assertNotEqual(event, event_diff)
+
+    def test_is_valid(self):
+        valid_event = app.Event(self.test_info)
+        self.assertTrue(valid_event.is_valid())
+
+        del self.test_info['name']
+        invalid_event = app.Event(self.test_info)
+        self.assertFalse(invalid_event.is_valid())
```
evankanderson: Add a test for behavior when at the arguments to `app.Event` contain a key not in attributes.

pageserve/app.py
```diff
@@ -22,24 +22,67 @@
 
 app = Flask(__name__)
 
-@app.route('/')
+
+@app.route('/v1/')
```
evankanderson: I'd recommend moving this change to a separate PR, since it looks like it should be unrelated to the events service.

pageserve/test.py
```diff
@@ -16,11 +16,20 @@
...
 class TestServe(unittest.TestCase):
-    def test_get_auth(self):
-        self.assertTrue(app.get_auth("carolyn"))
-        self.assertFalse(app.get_auth("Voldemort"))
-        
-        
+    def test_auth_json(self):
+        response = app.get_auth_json('example_user')
+        # edit_access should be either True or False
+        valid_response = response['edit_access'] or not response['edit_access']
```
evankanderson: Eventually, you should create two tests, one that tests yes-auth, and one that tests no-auth case.

events/app.py
cmei4444: done, switched to his method which is needed for testing with mongomock

events/app.py
```diff
@@ -0,0 +1,119 @@
...
+
+@app.route('/v1/add', methods=['POST'])
+def add_event():
+    # TODO(cmei4444): Get event information from request data
+    info = request.form['info']
+    # Add information into database
+    current_time = datetime.datetime.now()
+    build_event_info(info, current_time)
+    event = Event(info)
+    if not event.is_valid():
```
evankanderson: Why not have the Event constructor throw an exception if the supplied info is not valid?

events/app.py
evankanderson: Should this be required_attributes, and you accept other fields as well, or is this supposed to be all the attributes allowed?

events/app.py
```diff
@@ -0,0 +1,119 @@
...
+    pass
+
+
+@app.route('/v1/<event_id>', methods=['PUT'])
+def get_one_event(event_id):
+    """Retrieve one event by event_id"""
+    pass
+
+
+class Event(object):
```
evankanderson: Looking at this again and remembering python, would it be simpler to implement this using [`collections.namedtuple`](https://docs.python.org/2/library/collections.html#collections.namedtuple)?

pageserve/app.py
cmei4444: I'm having trouble figuring out how to move changes from past commits to different PRs, is there a good approach to this?

events/app.py
cmei4444: Took out the comment, since it's clear where the add to db is happening.

events/test.py
cmei4444: Took out the comment since the dict names are self-explanatory.

events/app.py
mukobi: Actually, the previous way I had of connecting to the db in if __name__ == '__main__' will break in Travi-CI, but you aren't seeing that because our test harness isn't set up to test the events directory. I've just gone on my branch and fixed up the database connecting to be all safe for testing, see the latest version at https://github.com/knative-portability/large-events/blob/0a95f7c496bcddbde80124069fd76e7c48ada800/users/app.py

events/test.py
cmei4444: Made this change.

events/app.py
cmei4444: I set up the db in a similar way to your current method, and set up testing for the events directory. Things seem to be working at the moment

mukobi: I actually just update things (sorry, I think this is the last one) to be simpler and not use the `global` keyword (which actually shows up as a warning from pylint). If you want I could switch to this later in a separate PR though. https://github.com/mukobi/large-events/blob/users-mongodb-atlas/users/app.py

pageserve/app.py
mukobi: Maybe you could make a new branch from this branch (which will carry over all of its commits) then revert all the commits on each branch that shouldn't be on their respective branches if that makes sense.

events/test.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class TestEventsClass(unittest.TestCase):
+    def setUp(self):
+        self.test_info = {'name': 'test_event',
+                          'description': 'testing!',
+                          'author': 'admin',
+                          'event_time': '7-12-2019',
+                          'created_at': '7-10-2019'}
+        self.test_info_with_id = {'name': 'test_event',
```
evankanderson: Consider using a copy here, like so:

```python
self.test_info_with_id = dict(event_id=1, **self.test_info)  # Only works if 'event_id' is not a key in self.test_info
```

pageserve/app.py
evankanderson: Yes, that's one way to do it.

If you're careful about your commits, you can use `git cherrypick` to copy over the commits. But I'm guessing that you don't really care about the commit history, so you can just:

```shell
git branch pr1  # Makes a new branch with the same commits as your current branch
git checkout upstream/master -- events # Reset the events files to what's in upstream/master
git commit -a -m "Reset events"
git checkout master  # Back to this branch
git checkout upstream/master -- pageserve  # Reset the pageserve files to what's in upstream/master
git commit -a -m "Reset pageserve"
git push origin
```

.travis.yml
```diff
@@ -22,5 +22,7 @@ env:
     USER_ENDPOINT=https://users-l3zyoxchqa-uc.a.run.app/v1/authorization
   - TEST_DIR=users
     USER_ENDPOINT=https://users-l3zyoxchqa-uc.a.run.app/v1/authorization
+  - TEST_DIR=events
+    USER_ENDPOINT=https://users-l3zyoxchqa-uc.a.run.app/v1/authorization
```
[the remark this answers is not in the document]
cmei4444: I think I'll need it for authorizing editing/creating events at some point so I included it for now

events/test.py
```diff
@@ -0,0 +1,89 @@
...
+                     'event_time': '7-12-2019', }
+        time = datetime.datetime(2017, 8, 28, 10, 33, 1, 100000)
+        info = app.build_event_info(test_info, time)
+        self.assertEqual(info['created_at'], time)
+
+        time_unrounded = datetime.datetime(2017, 8, 28, 10, 33, 1, 100654)
+        info = app.build_event_info(test_info, time_unrounded)
+        self.assertEqual(info['created_at'], time)
+
+    def tearDown(self):
```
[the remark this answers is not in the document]
cmei4444: Removed these entirely by deleting the DB between tests

events/app.py
cmei4444: I changed the functions for connecting to the DB to be very similar now.

events/app.py
cmei4444: The creation time should always be set when the POST request to create the event is received, which is why there isn't a default/overwrite.

events/app.py
```diff
@@ -88,40 +89,56 @@ def get_one_event(event_id):
...
+                pass
         return True
 
     def add_to_db(self, events_collection):
         """Adds the event to the specified collection."""
-        events_collection.insert_one(self.info)
+        events_collection.insert_one(self.get_info())
         return self.info
 
+    def get_info(self):
```
evankanderson: Consider using @property here to make this a read property, rather than as a method.

In particular, if you make Event hold all the data directly, then the "info as dict" version could be called `.info` or `.dict` or the like.

events/app.py
mukobi: You might try something like
`python
>>> import datetime
>>> datetime.datetime.now().isoformat()
'2019-07-15T13:20:18.085627'
`

events/test.py
cmei4444: Removed since this function no longer does rounding

events/app.py
cmei4444: Constructing the namedtuple causes this exception now

events/test.py
cmei4444: Done in  [9e98df4](https://github.com/knative-portability/large-events/pull/4/commits/9e98df4d5f973dfcdc86056ef61e4065c2e1bdc9)

events/app.py
cmei4444: Do we want to standardize how we represent dates/times across the app? I think ISO format can be hard to parse, so something like `'%Y-%m-%d %H:%M:%S'` would be better, and to convert back and forth we'd use `strftime` and `strptime`

mukobi: That sounds like a good idea, I like your format.

events/app.py
```diff
@@ -0,0 +1,154 @@
...
+    Used for comparing Event objects, since MongoDB truncates times to the
+    nearest millisecond.
+    """
+    if isinstance(time_1, type(time_2)):
+        if isinstance(time_1, datetime.datetime):
+            time_1 = time_1.replace(
+                microsecond=(time_1.microsecond // 1000) * 1000)
+            time_2 = time_2.replace(
+                microsecond=(time_2.microsecond // 1000) * 1000)
+        return time_1 == time_2
```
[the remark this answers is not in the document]
mukobi: Wow, that's so much cleaner! The more you know.

events/app.py
```diff
@@ -0,0 +1,154 @@
...
+            self = super(Event, cls).__new__(cls, **info)
+            return self
+        except TypeError:
+            raise ValueError("Event info was formatted incorrectly.")
+
+    def __eq__(self, other):
+        """Determines if two events have the same info, excluding event_id."""
+        if not isinstance(other, Event):
+            return False
+        for att in EVENT_ATTRIBUTES:
```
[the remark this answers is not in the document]
cmei4444: Yes, since an event is restricted to only having those fields so there shouldn't be any others.

events/app.py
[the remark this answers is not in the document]
cmei4444: It looks like `//` is actually integer division on timedeltas (from the link you included) so this seems to not work.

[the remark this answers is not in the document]
mukobi: That's unfortunate. Why again do we have to round to the nearest millisecond before comparison? Are you expecting the same event to be recorded in multiple places with times nanoseconds apart but that should be considered the same?

events/app.py
```diff
@@ -88,40 +89,56 @@ def get_one_event(event_id):
...
+    'name',
+    'description',
+    'author',
+    'created_at',
+    'event_time']
+
+EventTuple = namedtuple('EventTuple', attributes)
+
 
+class Event(object):
```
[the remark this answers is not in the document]
cmei4444: Implemented in [c36b19b](https://github.com/knative-portability/large-events/pull/4/commits/c36b19b2a180156208483f933359ab1e85e8013b)

events/app.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+    return Response(
+        status=201,
+    )
+
+
+def build_event_info(info, time):
+    """Adds created_at time to event info dict."""
+    info['created_at'] = time
+    return info
```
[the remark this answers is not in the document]
cmei4444: Done in [10703b5](https://github.com/knative-portability/large-events/pull/4/commits/10703b581ed58e4bcb9159704e48834f90275c93)

events/app.py
```diff
@@ -0,0 +1,154 @@
...
+@app.route('/v1/<event_id>', methods=['PUT'])
+def get_one_event(event_id):
+    """Retrieve one event by event_id."""
+    pass
+
+
+def equal_times(time_1, time_2):
+    """Determines if two times are equal.
+
+    Times can be represented as strings (mainly for testing) or as
```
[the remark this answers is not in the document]
cmei4444: Changed to all be datetime

events/app.py
```diff
@@ -0,0 +1,154 @@
...
+            return self
+        except TypeError:
+            raise ValueError("Event info was formatted incorrectly.")
+
+    def __eq__(self, other):
+        """Determines if two events have the same info, excluding event_id."""
+        if not isinstance(other, Event):
+            return False
+        for att in EVENT_ATTRIBUTES:
+            if att == 'event_time' or att == 'created_at':
```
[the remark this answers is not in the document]
cmei4444: Done in [10703b5](https://github.com/knative-portability/large-events/pull/4/commits/10703b581ed58e4bcb9159704e48834f90275c93)

events/app.py
```diff
@@ -0,0 +1,154 @@
...
+                if not equal_times(getattr(self, att), getattr(other, att)):
+                    print('times not eq')
+                    return False
+            elif att != 'event_id' and (
```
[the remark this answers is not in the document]
cmei4444: Done in [10703b5](https://github.com/knative-portability/large-events/pull/4/commits/10703b581ed58e4bcb9159704e48834f90275c93)

events/app.py
cmei4444: Done in [10703b5](https://github.com/knative-portability/large-events/pull/4/commits/10703b581ed58e4bcb9159704e48834f90275c93)

events/app.py
```diff
@@ -53,8 +53,7 @@ def add_event():
 
 def build_event_info(info, time):
     """Adds created_at time to event info dict."""
-    info['created_at'] = time
-    return info
+    return dict(created_at=time, **info)
```
mukobi: `info` comes from `request.form` which is user supplied, if I'm correct. How are you handling the case where `info` already has a `created_at` attribute? I think this dict constructor might raise a `KeyError` in that case.

events/test.py
```diff
@@ -40,21 +41,21 @@ def setUp(self):
...
-                          'event_time': '7-12-2019',
-                          'created_at': '7-10-2019'}
+                          'event_time': example_time,
+                          'created_at': example_time}
         self.test_info_with_id = dict(event_id=1, **self.test_info)
         self.test_info_with_db_id = dict(event_id=1, _id=2, **self.test_info)
 
     def test_construct_event(self):
         event = app.Event(self.test_info)
         self.test_info['event_id'] = None
```
mukobi: Doesn't your `Event` constructor automatically set `'event_id'` to `None` if it's not in the info?

events/app.py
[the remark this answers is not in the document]
evankanderson: It appears that mongodb stores millisecond-level precision, but the default for python is microsecond-level. This means that the microseconds number will be "wrong" (not-equal) after a round-trip through the database.

Unfortunately, it looks like @cmei4444 is correct, and the `//` operator only works on timedeltas (and returns an integer when it does so).

This looks about as clean as any other option; if we need to use this more, we should probably create a `monogodb_time` function to strip the milliseconds.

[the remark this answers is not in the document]
mukobi: I expect we will also need this for the posts service, so to follow DRY we should probably make some sort of common library for the whole project then.

events/app.py
evankanderson: Actually, you get a `TypeError`, surprisingly.

I just discovered [PEP 448](https://www.python.org/dev/peps/pep-0448/), which allows this form:

```python
return {**info, 'created_at': time}
```

events/app.py
```diff
@@ -0,0 +1,150 @@
...
+    current_time = datetime.datetime.now()
+    info = build_event_info(info, current_time)
+    try:
+        event = Event(info)
+    except ValueError:      # missing or extra event attributes
+        return Response(
+            status=400,
+            response="Event info was entered incorrectly.",
+        )
+    if EVENTS_COLL is not None:
```
evankanderson: If you have EVENTS_COLL be a thing that throws, you can restructure this to:

```python
try:
  event.add_to_db(EVENTS_COL)
  return Response(status=201)
except Exception as e:
  print("Oh no, an exception happened")
  traceback.print_exc()
  return Response(status=500, response="Oops")

File: events/app.py [hunk 32]
```diff
@@ -0,0 +1,150 @@
...
+    'description',
+    'author',
+    'created_at',
+    'event_time']
+
+
+class Event(namedtuple("EventTuple", EVENT_ATTRIBUTES)):
+    # TODO(cmei4444): move Event class to a separate file for better
+    # code organization
+    def __new__(cls, info):
```
evankanderson: You might want this to be `**info` if you want to be able to construct a new Event using keyword arguments, rather than needing to construct a dict. (You can always use the unpacking format to take a dict and turn it into keyword arguments.)

events/app.py
```diff
@@ -0,0 +1,150 @@
...
+        attributes are included.
+        """
+        if "event_id" not in info:
+            info['event_id'] = None
+        if '_id' in info:       # found DB-generated ID, override given one
+            info['event_id'] = info['_id']
+            del info['_id']
+        try:
+            self = super(Event, cls).__new__(cls, **info)
+            return self
```
evankanderson: You don't need the `self` temporary here, you can just `return super(...`

events/test.py
```diff
@@ -0,0 +1,118 @@
+import unittest
+import datetime
+import app
+import mongomock
+
+
+example_time = datetime.datetime(2019, 6, 11, 10, 33, 1, 100000)
```
evankanderson: Make this CONSTANT_CASE?

events/test.py
```diff
@@ -0,0 +1,118 @@
...
+        event_with_db_id = app.Event(self.test_info_with_db_id)
+        self.assertEqual(event_with_db_id.dict, self.test_info_with_db_id)
+
+    def test_constuct_event_error(self):
+        info_missing_name = self.test_info.copy()
+        del info_missing_name['name']
+        info_extra_field = self.test_info.copy()
+        info_extra_field['extra'] = 'hello'
+
+        self.assertRaises(ValueError, app.Event, info_missing_name)
```
evankanderson: Prefer the context manager version of assertRaises, as it's easier to read the code (and you can include multiple lines if needed:

```python
with self.assertRaises(ValueError):
  app.Event(info_missing_name)
```

events/app.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def connect_to_mongodb():
+    """Connects to MongoDB Atlas database.
+
+    Returns events collection if connection is successful, and None otherwise.
+    """
+    mongodb_uri = os.environ.get("MONGODB_URI")
+    if mongodb_uri is None:
+        print("Alert: not able to find MONGODB_URI environmental variable, "
```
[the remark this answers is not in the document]
mukobi: See https://github.com/knative-portability/large-events/pull/9/commits/92f1740d2eb916d45ba152a9c91775d200768f30 for my implementation

events/test.py
mukobi: `pylint -v app.py` should warn you against this.

events/app.py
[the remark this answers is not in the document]
cmei4444: A different route we could take would be to store all the times as strings in some agreed-upon format (I suggested one in an earlier comment), which is pretty easy to convert back and forth to datetimes

events/test.py
cmei4444: I wanted to test cases when event info had been queried, possibly after being re-entered, from the DB (so it would have an ID defined) in addition to when it was being user-supplied.

events/app.py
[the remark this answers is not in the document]
mukobi: > A different route we could take would be to store all the times as strings in some agreed-upon format (I suggested one in an earlier comment), which is pretty easy to convert back and forth to datetimes

I like that idea a lot better than all this datetime madness, especially because if we generate the string like you suggested earlier we can just not include the microsecond part and easily truncate it. I imaging the 2 things we really want these date times recorded for is to order events/posts by time and to give a client-side label on events/posts for when they were posted, both of which are easily accomplished by strings.

events/test.py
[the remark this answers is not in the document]
mukobi: `setUp` and `tearDown` run before and after every single test function with the test class (took me a while to figure this out). Your `setUp` created a new `mongomock.MongoClient` each time, so I don't think your `tearDown` is needed still.

events/app.py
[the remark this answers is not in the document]
evankanderson: Zeroing the microseconds value would also work here; if you decide on a string format, you probably want to:

1. Define the time zone that the times are in (you probably want UTC)
2. Ensure that times are consistently ascending (i.e. as the ISO timestamp format ensures)

events/app.py
evankanderson: (The reason for doing this would be to also handle exceptions from e.g. database writes.)

events/test.py
mukobi: Sorry, I'm a little confused. So for this first assert from lines 50-52, you're comparing `test_info` with an `Event` constructed from `test_info`, right? It seems to me that setting `self.test_info['event_id'] = None` shouldn't be necessary since your `Event.__eq__()` ignores `event_id`, no?

cmei4444: I'm actually comparing the info stored by the event object constructed from `self.test_info` with `self.test_info` itself. Since the constructor adds sets `event_id = None`, I have to do the same thing for `self.test_info`

events/app.py
cmei4444: Done in [d1ade21](https://github.com/knative-portability/large-events/pull/4/commits/d1ade21683ef7f0d5d8c52a230d650691c86cd9f)

events/app.py
cmei4444: Done in [d1ade21](https://github.com/knative-portability/large-events/pull/4/commits/d1ade21683ef7f0d5d8c52a230d650691c86cd9f)

events/test.py
cmei4444: Done in https://github.com/knative-portability/large-events/pull/4/commits/fb026b7885c5f190e191046cdac7c02cbfa72f9d

events/app.py
cmei4444: Done in https://github.com/knative-portability/large-events/pull/4/commits/ccab3b6f737717ba360d5ef9fb156313f751ead1

events/app.py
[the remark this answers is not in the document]
cmei4444: Done in https://github.com/knative-portability/large-events/pull/4/commits/ccab3b6f737717ba360d5ef9fb156313f751ead1

events/test.py
cmei4444: Done in https://github.com/knative-portability/large-events/pull/4/commits/d1ade21683ef7f0d5d8c52a230d650691c86cd9f

events/test.py
mukobi: Ah got it now, sorry for my confusion.

events/test.py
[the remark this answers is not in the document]
cmei4444: Deleted tearDown in https://github.com/knative-portability/large-events/pull/4/commits/eccdda43ca492244f830191beeecba72cfa7cca7

PR #8, opened 2019-07-15
Upsert user in users db

mukobi:
Add helper function to allow users service to update/insert a user object (e.g. from OAuth 2.0) in the users db along with corresponding tests.


users/app.py
```diff
@@ -54,15 +54,47 @@ def add_update_user():
     return jsonify(user_object), (201 if added_new_user else 200)
 
 
+def upsert_user_in_db(user_object, users_collection):
+    """Updates or inserts the user object into user_collection.
```
evankanderson: `users_collection` (lost an "s")

users/app.py
```diff
@@ -54,15 +54,47 @@ def add_update_user():
...
+
+    Adds the is_oganizer field to user_object with default value False.
```
evankanderson: `is_organizer` -- any reason to call out this field in particular? Do you want to simply say "Sets defaults if needed"?

users/app.py
```diff
@@ -54,15 +54,47 @@ def add_update_user():
...
+    Documents in users_collection should look like
+        {"user_id": foobar,
+         "name": Foo Bar,
+         "is_organizer": False}
+
+    Returns:
+        bool: whether user_object was formatted correctly and the upsert
+            was successful.
+    """
+    if "user_id" not in user_object or "name" not in user_object:
```
evankanderson: Consider using the dict view returned by `.keys()` to compare with a set of fields:

```python
if user_object.keys() != set("user_id", "name"):
  return False  # Why not throw?
```

users/app.py
```diff
@@ -54,15 +54,47 @@ def add_update_user():
...
+    user_object should look like
+        {"user_id": foobar,
+         "name": Foo Bar}
+
+    Documents in users_collection should look like
+        {"user_id": foobar,
+         "name": Foo Bar,
+         "is_organizer": False}
+
+    Returns:
```
evankanderson: Why not throw an AssertionError instead? Doesn't missing fields/etc mean that the object was malformed?

In some cases, it can be useful to return the created object if you have it on hand, since you've filled in additional fields on the object.

users/test.py
```diff
@@ -21,7 +21,7 @@
 
 import unittest
 import mongomock
-from app import find_authorization_in_db
+from app import find_authorization_in_db, upsert_user_in_db
```
evankanderson: I'd have a general preference for the form:

```python
import app

# ...
app.find_authorization_in_db(...)
```

As this gives the reader a little bit more of a hint that the function has been imported from another module.

users/test.py
```diff
@@ -67,5 +67,101 @@ def test_unkown_user_not_authorized(self):
             MALFORMATTED_IN_DB_USER, self.mock_collection))
 
 
+USER_ID = "valid-user-id"
+USER_NAME = "Valid User Name"
+ADDITIONAL_INFORMATION = "I'm not found in other documents"
```
evankanderson: Put all the constants together near the top of the file.

users/app.py
evankanderson: Reading your tests below, it sounds like you would want something like:

```python
if user_object.keys() <= set("user_id", "name"):
```

mukobi: Works with
```python
user_object.keys() < {"user_id", "name"}
```
Fixed in https://github.com/knative-portability/large-events/pull/8/commits/25039520f28fc4f4ce9d5194d2004f2499ab2ef6

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/fbb02301744bcc7b5b47d57687ab4f1184a31368

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/aa11d3de295c56d34f9d1a2853dc97c822ec2032

users/app.py
mukobi: Raising AssertionError done https://github.com/knative-portability/large-events/pull/8/commits/5b0030e9b87cac0f799e170a1ef0c68d074bb520

Returning created object done https://github.com/knative-portability/large-events/pull/8/commits/829984772e7d09ad55fb674aaf4234d2cb04dda2

users/test.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/14081bb4f73065e4ad403d07e73fc9e25e781250

users/test.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/69a2afea847898f941de63007c5a6a2db7ac0c0b

users/app.py
```diff
@@ -54,22 +54,54 @@ def add_update_user():
...
+    if user_object.keys() < {"user_id", "name"}:
+        raise AttributeError("missing user_id or name")
+    # insert is_organizer field (default False)
+    user_object["is_organizer"] = False
+    # upsert user in db
+    users_collection.update_one(
+        {"user_id": user_object["user_id"]},
+        {"$set": user_object},
+        upsert=True)
+    return users_collection.find_one({"user_id": user_object["user_id"]})
```
evankanderson: Looking at the Mongo API, it looks like `update_one` returns an [UpdateResult](https://api.mongodb.com/python/current/api/pymongo/results.html#pymongo.results.UpdateResult), which inculdes an `upserted_id` and a `raw_result`, which is "the raw result document returned by the server".

users/app.py
```diff
@@ -54,22 +54,54 @@ def add_update_user():
     return jsonify(user_object), (201 if added_new_user else 200)
 
 
+def upsert_user_in_db(user_object, users_collection):
+    """Updates or inserts the user object into users_collection.
+
+    Sets/overwrites default values as needed.
+
+    user_object should look like
```
evankanderson: I would say:

> `user_object` must contain a `user_id` and `name`, and may contain other attributes. Default values will be set as needed.

users/app.py
```diff
@@ -54,22 +54,54 @@ def add_update_user():
...
+        {"user_id": foobar,
+         "name": Foo Bar}
+
+    Documents in users_collection should look like
```
evankanderson: I'm not sure you want to document the schema here; if you do so, you'll need to update this comment if you change the schema in the future.

users/test.py
```diff
@@ -36,36 +40,139 @@ def setUp(self):
...
+
+    def test_insert_valid_user(self):
+        """A valid user object should be upserted and retrieved correctly."""
+        self.mock_collection = mongomock.MongoClient().db.collection
+        user_to_insert = {
+            "user_id": USER_ID,
+            "name": USER_NAME
+        }
+        app.upsert_user_in_db(user_to_insert, self.mock_collection)
+        found_user = self.mock_collection.find_one({"user_id": USER_ID})
```
evankanderson: You should check the return value of `app.upsert_user_in_db` and make sure the `user_id` matches from the returned object.

users/test.py
```diff
@@ -36,36 +40,139 @@ def setUp(self):
...
+    def test_malformatted_user_not_inserted(self):
+        """A malformatted user_object should not be inserted.
+
+        Note: app.upsert_user_in_db should not insert the user if it is missing
+        the 'user_id' or 'name' attributes, but should insert the user if
+        the user has more attributes than those. If the user is malformatted,
+        app.upsert_user_in_db raises an AttributeError, else it upserts the
+        user and returns the new user object.
+        """
+        self.mock_collection = mongomock.MongoClient().db.collection
```
evankanderson: You might want to initialize this in `setUp`?

users/test.py
```diff
@@ -36,36 +40,139 @@ def setUp(self):
...
+        # missing name
+        self.assertRaises(AttributeError, app.upsert_user_in_db,
```
evankanderson: I have a slight preference for the context manager format, as it's easier to see the invocation of the code:

```python
with self.assertRaises(AttributeError):
  app.upsert_user_in_db({"user_id": USER_ID}, self.mock_collection)
```

users/test.py
```diff
@@ -36,36 +40,139 @@ def setUp(self):
...
+        # upsert many times
+        for i in range(0, 42):
+            app.upsert_user_in_db(user_to_insert, self.mock_collection)
+        # only 1 user has been inserted
+        self.assertEqual(self.mock_collection.count_documents({}), 1)
+        found_user = self.mock_collection.find_one({"user_id": USER_ID})
+        self.assertEqual(user_to_insert["user_id"], found_user["user_id"])
+        self.assertEqual(user_to_insert["name"], found_user["name"])
+
+    def test_returns_upserted_user(self):
```
evankanderson: I would roll this test into the basic "happy path" in `test_insert_valid_user`.

users/app.py
mukobi: The raw_result field confusingly doesn't actually give back the whole document that was updated. Instead, it looks like
```python
{'connectionId': 0, 'err': None, 'n': 1, 'nModified': 0, 'ok': 1, 'updatedExisting': False, 'upserted': ObjectId('5d2d025007...1bb77ad4')}
```
It seems like it only gives some information about the updating operation. Since updating in Mongo allows you to partially update a document, I'm guessing it doesn't return the whole document in the case where the document is very large and that would be expensive but updating only a few attributes is cheap. Should I continue as-is with the extra db call after to get the object again or not return the object and put it on the caller to query the db again if they need to?

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/89632bcea2b0d94f886c3a9f5b4d03cc940150e2

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/89632bcea2b0d94f886c3a9f5b4d03cc940150e2

users/app.py
evankanderson: Given the (potential) cost, I would probably return the `upserted_id`, which should make it cheap for the caller to find the document if needed.

mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/29bbf737717d46f7bbb64968e726fc6792c4dbb8

users/test.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/e88cb3cbf6afc45fc448a0d0eb76915db152da8d

users/test.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/e88cb3cbf6afc45fc448a0d0eb76915db152da8d

users/test.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/4a27bbae23fd046b48d07e2ebd5e75b342ca2c70

users/test.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/b086a18acbed167342101132fff780d87be3095d

users/test.py
```diff
@@ -36,36 +40,136 @@ def setUp(self):
...
+        with self.assertRaises(AttributeError):
+            app.upsert_user_in_db({"name": USER_NAME}, self.mock_collection)
+        # missing both
+        with self.assertRaises(AttributeError):
+            app.upsert_user_in_db({}, self.mock_collection)
+        # no users should have been inserted
+        self.assertEqual(self.mock_collection.count_documents({}), 0)
+
+    def test_additional_info_still_inserted(self):
+        """A user_object with additional information should still be inserted.
```
cmei4444: Do you ever plan to use any of the extra info? If not, might be best to not insert it in order to prevent too much extra data being stored.

mukobi: I was thinking it would be useful to easily add email or phone info later for notifications which might be in some but not all users, but I guess we don't need that now so maybe it's best left out. @evankanderson ?

evankanderson: You should agree whether you want to be strict or lenient here. 

It sounds like we should go the strict route to start with.

users/app.py
```diff
@@ -54,22 +54,50 @@ def add_update_user():
...
+        ObjectID: The ID of the upserted object from the db. This can be used
+            to find the object with collection.find_one(object_id).
+
+    Raises:
+        AttributeError: if user_object is malformatted.
+    """
+    if user_object.keys() < {"user_id", "name"}:
+        raise AttributeError("missing user_id or name")
+    # insert is_organizer field (default False)
+    user_object["is_organizer"] = False
```
evankanderson: You may want to make a copy of the object here so you aren't modifying the passed-in input.

users/test.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/80bbf7263589bdc41a637f599b31f281c314b6f5

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/8/commits/80bbf7263589bdc41a637f599b31f281c314b6f5

mukobi:
Just responded to  comments, ready for more review.

PR #9, opened 2019-07-16
Connect posts service to db

mukobi:
Title

### DIFF
--- a/posts/app.py
+++ b/posts/app.py
@@ -19,10 +19,10 @@
 # limitations under the License.
 
 import os
-
+import pymongo
 from flask import Flask, jsonify, request
 
-app = Flask(__name__)
+app = Flask(__name__)  # pylint: disable=invalid-name
 
 
 @app.route('/v1/add', methods=['POST'])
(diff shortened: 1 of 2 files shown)


posts/app.py
```diff
@@ -53,5 +53,17 @@ def get_all_posts_for_event(event_id):
...
 
 
-if __name__ == "__main__":
+def connect_to_mongodb():  # pragma: no cover
+    """Connect to MongoDB instance using env vars."""
+    mongodb_uri = os.environ.get("MONGODB_URI")
+    if mongodb_uri is None:
+        print("Alert: not able to find MONGODB_URI environmental variable, "
+              "no connection to MongoDB instance")
+        return None  # not able to find db config var
```
evankanderson: Returning `None` rather than throwing an exception will mean that the `DB` global might be `None`. Do you expect the rest of the code to handle that case, or will it simply move the error to elsewhere in the program?

mukobi: The reason I made it like that is so that unit tests running in Travis won't crash when the environment variable isn't set (i.e. when it's hidden in a PR Travis build). In effect, this should only be None in that specific case, but in the event a user forgot to set it in the Cloud Run config or export it for local development for example, it prints out that it wasn't able to connect to Mongo and the rest of the code will crash when trying to access `DB`. That was the best compromise between cleanliness and verbosity I could find, but maybe there's a better way.

evankanderson: One other option is to return an object that throws a clear error on attribute access. You can override `__getattribute__` like so:

```python
class Thrower(object):
  def __getattribute__(self):
    throw DBNotConnected()
```

(You can even declare this class in the `connect_to_mongodb` method if you want to hide it from prying eyes.)

mukobi: Done https://github.com/knative-portability/large-events/pull/9/commits/92f1740d2eb916d45ba152a9c91775d200768f30

posts/app.py
```diff
@@ -53,5 +53,26 @@ def get_all_posts_for_event(event_id):
...
+
+    class DBNotConnectedError(EnvironmentError):
+        """Raised when not able to connect to the db."""
+
+    class Thrower(object):
+        """Used to raise an exception on failed db connect."""
+
+        def __getattribute__(self, _):
+            raise DBNotConnectedError(
+                "Not able to find MONGODB_URI environmental variable")
```
evankanderson: Minor nit: this use of "environmental" is non-standard, though correct English usage.

I'd switch it to "environment variable" even though that's a noun-noun pairing.

mukobi: Thanks for the clarification, done https://github.com/knative-portability/large-events/pull/9/commits/30a2993b7ebb2ee852338f35ef65d5f97bdea760

PR #10, opened 2019-07-16
Refactor testing to test runner pattern

mukobi:
Change automated test setup (including Travis-CI) to use a test runner pattern. Travis runs coverage on one `test_runner.py` file in each directory which handles importing sibling unit test modules, loading them into a test runner, and running all of them. 

This allows for a cleaner codebase, shorter individual test files, the isolation of tests relevant to a particular module or function of an app into different files, and the ability to easily test only certain parts of an app by running those seperate files.

To demonstrate this, I also went along and split up the horribly long `/users/test.py` file which previously had 2 test case classes into 2 separate files, one for each class.

### DIFF
--- a/posts/test_runner.py
+++ b/posts/test_runner.py
@@ -19,6 +19,7 @@ To see the tests, look at the sibling test_*.py files."""
 # limitations under the License.
 
 import unittest
+import sys
 
 import test_posts
 
@@ -31,5 +32,11 @@ def generate_suite():
     return suite
 
 
+def run_test_suite():
+    """Runs the test suite and exits with correct status code."""
+    result = unittest.TextTestRunner().run(generate_suite())
+    sys.exit(not result.wasSuccessful())
+
+
 if __name__ == "__main__":
-    unittest.TextTestRunner().run(generate_suite())
+    run_test_suite()
--- a/posts/test_runner.py
+++ b/posts/test_runner.py
@@ -1,5 +1,7 @@
-"""Unit tests for posts service.
-"""
+"""Unit test runner for users service.
+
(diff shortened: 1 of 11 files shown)


users/test_runner.py
```diff
@@ -0,0 +1,44 @@
...
+import unittest
+import sys
+
+import test_authorization
+import test_user_upsertion
+
+
+def generate_suite():
+    """Generates and loads a test suite."""
+    loader = unittest.TestLoader()
```
evankanderson: Consider using [`loader.discover(.)`](https://docs.python.org/3/library/unittest.html#unittest.TestLoader.discover) rather than explicitly loading the modules.

`os.path.dirname(__file__)` may be helpful for finding the current file's directory.

mukobi: Thanks, makes things much cleaner. Done in https://github.com/knative-portability/large-events/pull/10/commits/8f14a6ce43a7e631cb9a26e0e97c80cc15d4174b

mukobi:
Addressed Evan's review, implementing a change that simplified things dramatically (now all `test_runner.py` files are the same except for their module docstrings). Also merged in Carolyn's changes from #4 and added a test runner and independent test files to `events`.

posts/test_runner.py
```diff
@@ -0,0 +1,33 @@
+"""Unit test runner for users service.
```
cmei4444: should be posts service, not users

mukobi: Thanks, fixed in https://github.com/knative-portability/large-events/pull/10/commits/a7636bc15b2d823226ae2e327b89bee1b3237fbf

PR #11, opened 2019-07-17
Users allow change authorization

mukobi:
Add a helper function to users service to allow only for the changing of authorization levels in the database and test it.

### DIFF
--- a/users/app.py
+++ b/users/app.py
@@ -78,7 +78,8 @@ def upsert_user_in_db(user_object, users_collection):
         upsert=True).upserted_id
 
 
-def update_user_authorization_in_db(user_id, is_organizer, users_collection):
+def update_user_authorization_in_db(
+        user_id: str, is_organizer: bool, users_collection):
     """Updates the authorization of the given user in the database.
 
     Assumes caller has authorization to make this change in the database.
@@ -95,13 +96,10 @@ def update_user_authorization_in_db(user_id, is_organizer, users_collection):
 
     Raises:
         KeyError: Bad `user_id`/user not found in db.
-        TypeError: `is_organizer` is not a bool.
     """
-    if not isinstance(is_organizer, bool):
-        raise TypeError("Trying to set authorization to a non-bool type.")
     result = users_collection.update_one(
         {"user_id": user_id},
-        {"$set": {"is_organizer": is_organizer}},
+        {"$set": {"is_organizer": bool(is_organizer)}},
         upsert=False)
     if result.matched_count == 0:
         raise KeyError(f"User with ID '{user_id}' not found.'")
--- a/users/app.py
+++ b/users/app.py
@@ -79,6 +79,24 @@ def upsert_user_in_db(user_object, users_collection):
 
 
 def update_user_authorization_in_db(user_id, is_organizer, users_collection):
+    """Updates the authorization of the given user in the database.
+
+    Assumes caller has authorization to make this change in the database.
+    I.e. caller route should first make sure request maker was authorized
+    to update the given user.
+
+    Args:
+        user_id (str): The id of the user to change.
+        is_organizer (bool): The authorization value to set.
+
+    Returns:
+        ObjectID: The ID of the updated object from the db. This can be used
+            to find the object with `collection.find_one(object_id)`.
+
+    Raises:
+        IndexError: Bad `user_id`/user not found in db.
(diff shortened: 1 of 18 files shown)


users/app.py
```diff
@@ -78,6 +78,36 @@ def upsert_user_in_db(user_object, users_collection):
...
+    Returns:
+        ObjectID: The ID of the updated object from the db. This can be used
+            to find the object with `collection.find_one(object_id)`.
+
+    Raises:
+        KeyError: Bad `user_id`/user not found in db.
+        TypeError: `is_organizer` is not a bool.
+    """
+    if not isinstance(is_organizer, bool):
+        raise TypeError("Trying to set authorization to a non-bool type.")
```
mukobi: Question: is this a good way to handle this? This helper function should only be used by 1 route in the same file later, so it probably will never be called incorrectly, so maybe this case isn't even needed. I wanted to type check `is_organizer` though so that it could never be entered as a non-bool value into the database. Would [PEP 484](https://www.python.org/dev/peps/pep-0484/) (type hints) or [PEP 3107](https://www.python.org/dev/peps/pep-3107/) (function annotations) be a better way to just hint at the correct type without strict code enforcement, or is this fine?

users/test_update_authorization.py
```diff
@@ -0,0 +1,112 @@
...
+            {"user_id": AUTHORIZED_USER})["is_organizer"])
+        # False to False
+        self.assertFalse(self.mock_collection.find_one(
+            {"user_id": NON_AUTHORIZED_USER})["is_organizer"])
+        app.update_user_authorization_in_db(
+            NON_AUTHORIZED_USER, False, self.mock_collection)
+        self.assertFalse(self.mock_collection.find_one(
+            {"user_id": NON_AUTHORIZED_USER})["is_organizer"])
+
+    def test_set_authorization_of_malformed_user(self):
```
cmei4444: Do you expect to have malformed users in the DB?

mukobi: Not really, but I wanted to make sure the behaviour is defined safely in that case. @evankanderson ?

users/app.py
evankanderson: PEP 484 and duck typing (e.g. `bool(is_organizer)`) should be sufficient here.

Overall, I'd recommend PEP 484 if you have time.

mukobi: Done https://github.com/knative-portability/large-events/pull/11/commits/a94e699754b72786210719bad71125fdc9b94f24 . Does that look right?

users/test_update_authorization.py
```diff
@@ -0,0 +1,112 @@
...
+        self.assertIsNone(self.mock_collection.find_one(
+            {"user_id": MISSING_USER}))
+
+    def test_fail_on_bad_arg_type(self):
+        """Should raise error and do nothing on bad arg type."""
+        self.assertTrue(self.mock_collection.find_one(
+            {"user_id": AUTHORIZED_USER})["is_organizer"])
+        with self.assertRaises(TypeError):
+            app.update_user_authorization_in_db(
+                AUTHORIZED_USER, "I am not a bool", self.mock_collection)
```
evankanderson: `"False"` might be a good value to test with here, because in python, `"False"` is a non-empty string, so `bool("False")` is `True`.

Using `"False"` as an example also suggests *why* you care about this case.

users/test_update_authorization.py
mukobi: Cool, thanks for the clarification.

users/test_update_authorization.py
mukobi: So in app.update_user_authorization_in_db should I still raise an error when the type is not bool like before, or should duck typing with `bool("False")` is `True` be the expected behaviour?

cmei4444: I think the error raising is still necessary - with this approach the code may behave confusingly (for example if is_organizer is a string) and it seems best to make sure that can't happen altogether

mukobi: Done https://github.com/knative-portability/large-events/pull/11/commits/3f19cef83509aca7fa05b0aa9f06f2c1f783d186

users/app.py
evankanderson: I think so -- you might want to check and see if you can get the type checking to alarm on passing the string `"False"` to `is_organizer`.

PR #13, opened 2019-07-18
Created some basic UI for pageserve

cmei4444:
Created basic web interface to display events and posts with some example data.
Most changes moved over from reverted changes on previous PR


pageserve/test_serve.py
```diff
@@ -18,9 +18,21 @@
 
 
 class TestServe(unittest.TestCase):
-    def test_get_auth(self):
-        # TODO: test authorization retrieval
-        pass
+    def test_auth_json(self):
```
cmei4444: So right now this test is more of an integration test between pageserve and users, which sometimes causes it to fail in Travis, any suggestions for restructuring the get_auth_json function in a better way?

evankanderson: Given the current structure, mocks with injection to replace the requests module are probably the best way to test this.

evankanderson: The other option would be to use something like https://github.com/pretenders/pretenders, which looks to let you set up your own HTTP server, but this part may need to be integration tested.

pageserve/test_serve.py
```diff
@@ -18,9 +18,21 @@
...
+        """Checks if users service returns a correctly formatted object.
+
+        A dictionary with a boolean 'edit_access' field should be received.
+        """
+        response = app.get_auth_json('example_user')
+        valid_response = response['edit_access'] or not response['edit_access']
+        self.assertTrue(valid_response)
```
mukobi: Right now this test kinda just makes sure you don't get an AttributeError trying to access 'edit_access'.  response.json() from your get_auth_json function decodes json into a Python dict, so response['edit_access'] should be guaranteed to be a bool. As it is now, if response['edit_access'] is something else like a string, I think your code implicitly casts it to a bool and would probably pass. To fix this, you could just add another assert to make sure response['edit_access'] is an instance of bool.

evankanderson: It seems like the defensive option would be to treat anything except `True` as `False`, using `response['edit_access'] is True` in `has_edit_access` (or rolling that into `get_auth_json` if you don't need to do anything else with the returned information from `get_auth_json`.

pageserve/app.py
```diff
@@ -22,24 +22,93 @@
...
         'index.html',
+        posts=posts,
         auth=is_auth,
     )
-    
-def get_auth(user):
+
+
+@app.route('/v1/events')
+def show_events():
```
mukobi: Function docstring?

pageserve/app.py
```diff
@@ -22,24 +22,93 @@
...
+    is_auth = has_edit_access(get_auth_json(user))
+    events = get_events()
+    return render_template(
+        'events.html',
+        events=events,
+        auth=is_auth,
+    )
+
+
+def get_posts():
```
mukobi: Function docstring?

pageserve/app.py
```diff
@@ -22,24 +22,93 @@
...
+              'event_id': '0',
+              'type': 'image',
+              'author': 'admin',
+              'created_at': '7-9-2019',
+              'text': 'abcdefghi',
+              }]
+    return parsed_posts(posts)
+
+
+def parsed_posts(posts):
```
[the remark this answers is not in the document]
mukobi: It looks like maybe you wanted to get post json with a request and then use this to parse the json, but if that's the case you should probably comment this with a docstring and TODO or else delete this function.

pageserve/app.py
```diff
@@ -22,24 +22,93 @@
...
+
+def get_posts():
+    # TODO: integrate with posts service to pull post info from database
+    posts = [{'post_id': '1',
+              'event_id': '0',
+              'type': 'text',
+              'author': 'admin',
+              'created_at': '7-9-2019',
+              'text': 'this will be a fun event!',
+              },
```
mukobi: Do you want to handle images yet or implement that later?

pageserve/app.py
```diff
@@ -22,24 +22,93 @@
...
+              'text': 'abcdefghi',
+              }]
+    return parsed_posts(posts)
+
+
+def parsed_posts(posts):
+    return posts
+
+
+def get_events():
```
mukobi: Function docstring?

pageserve/app.py
```diff
@@ -22,24 +22,93 @@
...
+               'event_time': '7-12-2019',
+               }]
+    return parsed_events(events)
+
+
+def parsed_events(events):
+    return events
+
+
+def get_auth_json(user):
```
mukobi: Function docstring?

pageserve/app.py
```diff
@@ -22,24 +22,93 @@
...
     url = os.environ.get("USER_ENDPOINT")
-    r = requests.post(url, data={'user_id':user})
+    r = requests.post(url, data={'user_id': user})
     response = r.json()
-    is_auth = response['edit_access']
-    return is_auth
+    return response
+
+
+def has_edit_access(user_data):
```
mukobi: Function docstring?

pageserve/app.py
```diff
@@ -22,24 +22,93 @@
...
+    return user_data['edit_access']
+
 
 def get_user():
```
mukobi: Function docstring?

pageserve/test_serve.py
mukobi: I think we wanted to make the test independent of whether the given user actually has edit access in the db that the users service is deployed with, so a better comparison might be 
```python
self.assertTrue(response['edit_access'] is True or response['edit_access'] is False)
```

pageserve/test_serve.py
cmei4444: Still a little confused what exactly I'd be testing with a mock - is it just that the get_auth_json goes through all the necessary steps? or would I still test the returned value?

pageserve/app.py
cmei4444: I'll still take images from the posts service but if I don't get to displaying them I'll just put a placeholder instead

pageserve/test_serve.py
mukobi: You could mock `requests` for the scope of this function and make `.get` return to give you what you want. Using `unittest.mock`, something like
```python
mock_requests = mock.MagicMock()
mock_requests.get().return_value = '{"edit_access":"true"}'
with mock.patch('app.requests', mock_requests):
    response = app.get_auth_json('example_user')
valid_response = response['edit_access'] or not response['edit_access']
self.assertTrue(valid_response)
```

evankanderson: You can write the context manager as:

```python
with mock.patch('app.requests') as mock_requests:
  mock_requests.get().return_value = '{"edit_access":"true"}'
  response = app.get_auth_json('example_user')
...
```

pageserve/test_serve.py
evankanderson: The test should know whether `edit_access` should be `True` or `False` in this test.

pageserve/test_serve.py
cmei4444: @evankanderson what's the best way to handle the r.json() function here? it's expecting to be called on a response from the requests library and I'm not sure how to patch both functions at once

mukobi:
```suggestion
with mock.patch('app.requests') as mock_requests:
    mock_response = mock.MagicMock()
    mock_response.json().return_value = '{"edit_access":"true"}'
    mock_requests.get().return_value = mock_response
    response = app.get_auth_json('example_user')
```

cmei4444: That did it, thanks!

cmei4444: I did have to take change json() to json - same for post().

pageserve/test_serve.py
cmei4444: Changed test to reflect this in e163db6e3d8bc96d12dd6611fec67bdd3bbff5b9

pageserve/app.py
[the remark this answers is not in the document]
cmei4444: Clarified with a TODO in 734eefb65a0b9e961c59c484e9d33c70387d8c66

mukobi:
Cool, looks pretty good. I noticed on [Coveralls](https://coveralls.io/builds/24757499) that our code coverage is really low, but pageserve has the second lowest at 53%. I'm currently working on figuring out how to properly unit test our flask routes, and I'm working on bringing up coverage on some of the other dirs. Maybe you could make a branch off of this one and improve some of the pageserve coverage (coveralls as well as `coverage report` show you which lines aren't covered). For how to test the Flask routes, see here: https://github.com/mukobi/large-events/blob/test-routes-improve-coverage/users/test_routes.py

evankanderson:
looks good!

PR #15, opened 2019-07-18
Move event class out of app.py

cmei4444:
Move class out of main file, for possible future use in other files and better organization.


events/app.py
```diff
@@ -47,7 +47,7 @@ def add_event():
             response="Event info was entered incorrectly.",
         )
     try:
-        event.add_to_db(EVENTS_COLL)
+        add_to_db(event.dict, EVENTS_COLL)
```
evankanderson: Given the availability of mongomock, why not have this just call `EVENTS_COLL.insert_one(event.dict)`?

cmei4444: Thanks - I didn't realize that it was simpler to just add to the DB directly there.

PR #16, opened 2019-07-18
Post upload helper

mukobi:
Helper function to allow for uploading a post into the posts db along with unit tests.

### DIFF
--- a/posts/app.py
+++ b/posts/app.py
@@ -79,7 +79,7 @@ def upload_new_post_to_db(post, collection):
     if post.keys() != required_attributes:
         raise AttributeError(f"Arg post must have exactly the "
                              "attributes {required_attributes}")
-    if str(post["text"]) == "" and list(post["files"]) == []:
+    if not post["text"] and not post["files"]:
         raise ValueError("One of text or files must not be empty.")
     # post is valid, add on timestamp and insert into db
     post["created_at"] = generate_timestamp()
--- a/posts/app.py
+++ b/posts/app.py
@@ -78,7 +78,19 @@ def upload_new_post_to_db(post, collection):
                              "attributes {required_attributes}")
     if str(post["text"]) == "" and list(post["files"]) == []:
         raise ValueError("One of text or files must not be empty.")
+    # post is valid, add on timestamp and insert into db
+    post["created_at"] = generate_timestamp()
     return collection.insert_one(post).inserted_id
+
+
+def generate_timestamp() -> str:
+    """Generate timestamp of the current time for placement in db.
+
+    Returns:
+        str: string representation of the current time.
+    """
+    # TODO use general timestamp generation function from events
+    return "2017-10-06T00:00:00+00:00"
 
 
 def connect_to_mongodb():  # pragma: no cover
--- a/posts/app.py
+++ b/posts/app.py
@@ -56,6 +56,9 @@ def get_all_posts_for_event(event_id):
 def upload_new_post_to_db(post, collection):
     """Uploads a new post to the db collection.
 
+    Assumes the event matching the post's `event_id` and the user matching
+        the post's `author_id` both exist. I.e. Caller should check this.
+
     Args:
         post (dict): Post to add. Requires exactly the following attributes:
             event_id (str): id of the event to post to
--- a/posts/app.py
+++ b/posts/app.py
@@ -54,7 +54,31 @@ def get_all_posts_for_event(event_id):
 
 
 def upload_new_post_to_db(post, collection):
-    """Uploads a new post to the db collection."""
+    """Uploads a new post to the db collection.
+
+    Args:
+        post (dict): Post to add. Requires exactly the following attributes:
+            event_id (str): id of the event to post to
+            author_id (str): user id of the user making the post
+            text (str): text description
+            files (list): list of string encoded files
+        collection: pymongo collection to insert into.
+
+    Returns:
+        ObjectID: DB ID of the post that was uploaded.
+
+    Raises:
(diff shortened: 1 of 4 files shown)


posts/app.py
```diff
@@ -53,6 +53,49 @@ def get_all_posts_for_event(event_id):
     """Get all posts matching the event with the specified ID."""
 
 
+def upload_new_post_to_db(post, collection):
+    """Uploads a new post to the db collection.
+
+    Assumes the event matching the post's `event_id` and the user matching
+        the post's `author_id` both exist. I.e. Caller should check this.
```
evankanderson: Is it important that both exist? What happens if they do not?

It seems like this function will behave fine if they don't exist?

posts/app.py
```diff
@@ -53,6 +53,49 @@ def get_all_posts_for_event(event_id):
...
+    Raises:
+        ValueError: Has no text body (i.e. empty string) nor any files
+            to upload.
+        AttributeError: `post` has not enough or too many attributes.
+    """
+    required_attributes = {"event_id", "author_id", "text", "files"}
+    if post.keys() != required_attributes:
+        raise AttributeError(f"Arg post must have exactly the "
+                             "attributes {required_attributes}")
+    if str(post["text"]) == "" and list(post["files"]) == []:
```
evankanderson: Can you use python truthiness here?

```python
if not post["text"] and not post["files"]:
  raise ...
```

posts/test_post_uploading.py
```diff
@@ -0,0 +1,126 @@
...
+    def setUp(self):
+        """Set up mock DB for testing"""
+        self.mock_collection = mongomock.MongoClient().db.collection
+
+    def test_full_upload(self):
+        """Can upload a full post object with both text and files."""
+        app.upload_new_post_to_db(VALID_POST_FULL, self.mock_collection)
+        post_in_db = self.mock_collection.find_one({})
+        self.assertIsNotNone(post_in_db)
+        self.assertTrue(self.posts_are_equal(VALID_POST_FULL, post_in_db))
```
evankanderson: Consider making `posts_are_equal` a classmethod like `assertPostsEqual()` which reports both values when they are not equal.

posts/test_post_uploading.py
```diff
@@ -0,0 +1,126 @@
...
+    def test_partial_upload(self):
+        """Can upload a post with one of text or files."""
+        # text but no files
+        app.upload_new_post_to_db(
+            VALID_POST_TEXT_NO_FILES, self.mock_collection)
+        post_in_db = self.mock_collection.find_one({})
+        self.assertIsNotNone(post_in_db)
+        self.assertTrue(self.posts_are_equal(
+            VALID_POST_TEXT_NO_FILES, post_in_db))
+        # files but no text
```
evankanderson: Is there any need for this to be part of the same test, rather than a new test?

posts/app.py
mukobi: It's important that they both exist so that the front end UI is able to query all posts for a given event ID and see the users that posted each post. Right now, this function behaves fine if they don't exist which is why I added this comment.

posts/app.py
mukobi: Yeah, I was overthinking it. Done https://github.com/knative-portability/large-events/pull/16/commits/9020e0fd725fd4c8fe9bb1d0695f6348e2b3936f

posts/test_post_uploading.py
mukobi: Done https://github.com/knative-portability/large-events/pull/16/commits/9020e0fd725fd4c8fe9bb1d0695f6348e2b3936f

posts/test_post_uploading.py
mukobi: No besides them doing similar things. I split them up in https://github.com/knative-portability/large-events/pull/16/commits/9020e0fd725fd4c8fe9bb1d0695f6348e2b3936f

mukobi:
Addressed Evan's review, ready for more.

posts/app.py
evankanderson: One part of microservice design is that you don't want to have one domain (microservice) worry too much about constraints in another domain (microservice). So in this case, it would be the frontend that managed the concern about handling missing events and/or authors.

posts/test_post_uploading.py
```diff
@@ -0,0 +1,124 @@
...
+        self.assertEqual(first["event_id"], second["event_id"])
+        self.assertEqual(first["author_id"], second["author_id"])
+        self.assertEqual(first["text"], second["text"])
+        self.assertEqual(first["files"], second["files"])
+
+    def setUp(self):
+        """Set up mock DB for testing"""
+        self.mock_collection = mongomock.MongoClient().db.collection
+
+    def test_full_upload(self):
```
evankanderson: You seem to have two cases here:

1. Post a certain payload, expect that object to exist.
2. Post a certain payload, expect no object to exist.

This would be a good use case for table-driven testing, but it appears that table-driven tests are harder to write in python than in Go:

http://tomlinford.com/post/table-driven-tests/ is one reference; these are sometimes also called "parameterized tests".

posts/app.py
mukobi: Ah okay, done in https://github.com/knative-portability/large-events/pull/16/commits/775e718d10cb96e0abf66da99e9c2b80c7f2882d.

posts/test_post_uploading.py
mukobi: So under table testing all these tests would be in a single function, correct? I'll look into implementing this later.

PR #18, opened 2019-07-18
Clean up travis (maybe fix coveralls)

mukobi:
Clean up .travis.yml and possibly fix coveralls reporting 0% coverage.


mukobi:
> Do you know what was causing the issue? It looks like coverage is still 0% in the tests

I still can't figure it out, but this PR also cleans up .travis.yml by removing an unused environment variable and unwrapping the &&s from the test script.

PR #19, opened 2019-07-19
Get all events

cmei4444:
Wrote function to respond to request to get all events stored in DB

### DIFF
--- a/events/app.py
+++ b/events/app.py
@@ -77,16 +77,9 @@ def get_all_events():
         events_dict = build_events_dict(events)
         # TODO(cmei4444): test with pageserve to make sure the json format is
         # correct in the response
-        return Response(
-            events_dict,
-            status=200,
-            mimetype='application/json'
-        )
+        return jsonify(events_dict)
     except DBNotConnectedError as e:
-        return Response(
-            status=500,
-            response="Database was undefined.",
-        )
+        return "Database was undefined.", 500
 
 
 def build_events_dict(events):
--- a/events/app.py
+++ b/events/app.py
@@ -2,7 +2,7 @@ import os
 import datetime
 import pymongo
 
-from flask import Flask, request, Response
+from flask import Flask, request, Response, json
 from eventclass import Event
 
 app = Flask(__name__)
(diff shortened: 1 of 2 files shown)


events/app.py
```diff
@@ -72,7 +72,31 @@ def edit_event(event_id):
...
+    try:
+        events = events_collection.find()
+        events_dict = build_events_dict(events)
+        # TODO(cmei4444): test with pageserve to make sure the json format is
+        # correct in the response
+        return Response(
+            events_dict,
+            status=200,
+            mimetype='application/json'
+        )
```
mukobi: Instead of importing `json` and manually constructing a response, you can import `jsonify` from Flask.
```python
return jsonify(events_dict)
```
Status code 200 is default.

events/app.py
```diff
@@ -72,7 +72,31 @@ def edit_event(event_id):
...
+    except DBNotConnectedError as e:
+        return Response(
+            status=500,
+            response="Database was undefined.",
+        )
```
mukobi: In flask you can also do
```python
return "Database was undefined.", 500
```
Which may be a bit cleaner. The documented use of the `Response` constructor seems to be if you need to do something fancy like changing headers, but you aren't yet so you can rely on implicit response creation.

events/test_events_db.py
```diff
@@ -3,12 +3,12 @@
 import mongomock
 import app
 
-
 EXAMPLE_TIME = datetime.datetime(2019, 6, 11, 10, 33, 1, 100000)
 
 
 class TestEventsDB(unittest.TestCase):
     def setUp(self):
+        self.maxDiff = None
```
mukobi: I'm not sure I see where this is currently used.

cmei4444: Oops! This was for testing and I forgot to remove it, thanks for catching it

events/app.py
```diff
@@ -72,7 +72,24 @@ def edit_event(event_id):
...
     """Return a list of all events currently in the DB."""
-    pass
+    try:
+        events = events_collection.find()
+        events_dict = build_events_dict(events)
+        # TODO(cmei4444): test with pageserve to make sure the json format is
+        # correct in the response
+        return jsonify(events_dict)
+    except DBNotConnectedError as e:
+        return "Database was undefined.", 500
```
whaught: Maybe include the database name in the error?

events/app.py
```diff
@@ -72,7 +72,24 @@ def edit_event(event_id):
 @app.route('/v1/', methods=['GET'])
 def get_all_events():
     """Return a list of all events currently in the DB."""
-    pass
+    try:
+        events = events_collection.find()
+        events_dict = build_events_dict(events)
```
evankanderson: Do you need this temporary? I'd write this without the helper as:

```python
try:
  events = events_collection.find()
  return jsonify({'num_events': len(events), 'events': events})
except DbNotConnectedError as e:
  return ...
```

PR #24, opened 2019-07-23
Posts serve from db

mukobi:
Add helper and endpoints to serve posts. Can query all posts (/v1/), a post matching a specific id (/v1/<post_id>), or all posts matching a given event id (/v1/by_event/<event_id>).


posts/app.py
```diff
@@ -60,28 +61,52 @@ def upload_new_post():
 @app.route('/v1/', methods=['GET'])
 def get_all_posts():
     """Get all posts for the whole event."""
-    # TODO(mukobi) change from fake data to real query from db
-    return jsonify({
```
whaught: Nice! glad to be rid of this fake static data

posts/app.py
```diff
@@ -60,28 +61,52 @@ def upload_new_post():
...
 @app.route('/v1/by_event/<event_id>', methods=['GET'])
 def get_all_posts_for_event(event_id):
     """Get all posts matching the event with the specified ID."""
+    # serialize otherwise nonserializable ObjectIDs
+    return json_util.dumps(
+        find_posts_in_db(POSTS_COLLECTION, event_id=event_id))
+
+
+def find_posts_in_db(collection, post_id=None, event_id=None):
+    """Finds all matching posts in the database.
```
whaught: Meta comment: these method comments are thorough and awesome.

posts/app.py
```diff
@@ -60,28 +61,52 @@ def upload_new_post():
...
-        "author_id": "mukobi",
-        "created_at": "2017-10-06T00:00:00+00:00",
-        "text": "This is the description of a post. The post is tagged with "
-                "the event id of 'all' and should be viewable by default.",
-        "images": [
-            "https://upload.wikimedia.org/wikipedia/commons/6/66/An_up-close_picture_of_a_curious_male_domestic_shorthair_tabby_cat.jpg",
-            "http://4.bp.blogspot.com/-w8U75TCuhgU/Tzw8TmaclvI/AAAAAAAABJ0/6fMMcRLAceM/s1600/Rabbit3.jpg"
-        ]})
+    # serialize otherwise nonserializable ObjectIDs
+    return json_util.dumps(find_posts_in_db(POSTS_COLLECTION))
```
cmei4444: Do you need to load this into json again? At least for the response type to be recognized as json, not sure if it's automatic.

posts/app.py
mukobi: There goes our fake reliability

posts/app.py
mukobi: Thanks!

posts/app.py
mukobi: Thanks, done in https://github.com/knative-portability/large-events/pull/24/commits/4000202c4c42a327ab14761ea5c3ad6a97f79d96 https://github.com/knative-portability/large-events/pull/24/commits/c67368b4aec6bf6aa196edf45f3a1f21527b2972 where previously it was returning text. I also had to wrap the list result of the query into a dict for valid json, so now responses look like ```
{javascript
    "posts": {
        [
            post1...,
            post2...
        ]
    },
    "num_posts": 2
}
```
as per the feature doc.

File: posts/app.py [hunk 4]
```diff
@@ -60,28 +62,59 @@ def upload_new_post():
...
-        "created_at": "2017-10-06T00:00:00+00:00",
-        "text": "This is the description of a post. The post is tagged with "
-                "the event id of 'all' and should be viewable by default.",
-        "images": [
-            "https://upload.wikimedia.org/wikipedia/commons/6/66/An_up-close_picture_of_a_curious_male_domestic_shorthair_tabby_cat.jpg",
-            "http://4.bp.blogspot.com/-w8U75TCuhgU/Tzw8TmaclvI/AAAAAAAABJ0/6fMMcRLAceM/s1600/Rabbit3.jpg"
-        ]})
+    # serialize otherwise nonserializable ObjectIDs
+    post_list = find_posts_in_db(POSTS_COLLECTION)
+    return json.loads(json_util.dumps(
```
cmei4444: Consider changing this into a function so you don't have to repeat it

posts/test_post_serving.py
```diff
@@ -0,0 +1,106 @@
...
+        self.collection.insert_many(FAKE_POSTS)
+
+    def test_find_all_posts(self):
+        """Find all fake posts."""
+        found = app.find_posts_in_db(self.collection)
+        self.assertEqual(found, FAKE_POSTS)
+
+    def test_find_by_post_id(self):
+        """Search for 1 post by post ID."""
+        post_id = FAKE_POSTS[0]["_id"]
```
cmei4444: I don't see where an "_id" attribute is defined in FAKE_POSTS, where does it come from?

mukobi: I just found this out today: `collection.insert_many(document)` will actually add a `"_id"` attribute to `document` to match the ID that is generated in the MongoDB collection if you didn't already add a `"_id"` to `document`, which is pretty cool. I found this out when a test comparing the fake object to the one in the db said they were the same which I was not expecting. I've gone ahead an added a comment explaining this https://github.com/knative-portability/large-events/pull/24/commits/de1938ec530ea005e406717d092602c296daf6ea

posts/app.py
mukobi: Done, thanks for DRY reminder https://github.com/knative-portability/large-events/pull/24/commits/de1938ec530ea005e406717d092602c296daf6ea

PR #26, opened 2019-07-24
Call events service from pageserve

cmei4444:
Retrieves and displays all events in the database in the events.html template.


pageserve/app.py
```diff
@@ -76,29 +76,19 @@ def parsed_posts(posts):
...
-               'author': 'admin',
-               'created_at': '7-9-2019',
-               'event_time': '7-12-2019',
-               }]
-    return parsed_events(events)
-
-
-def parsed_events(events):
+    url = os.environ.get("EVENTS_ENDPOINT")
+    r = requests.get(url, params={})
```
mukobi: Does this work? It look like $EVENTS_ENDPOINT in Travis is missing the /v1/

pageserve/app.py
```diff
@@ -76,29 +76,19 @@ def parsed_posts(posts):
...
+    if r.status_code == 200:
+        return parsed_events(r.json())
+    else:
+        # TODO(cmei4444): handle error in a way that doesn't break page display
+        return "Error in getting events"
+
+
+def parsed_events(events_dict):
```
mukobi: Add docstring to explain what this does and maybe why you need to extract the `'events'` list from `events_dict`.

.travis.yml
```diff
@@ -21,6 +21,7 @@ env:
   # Travis build matrix will test all permutations of the build
   - TEST_DIR=pageserve
     USER_ENDPOINT=https://users-l3zyoxchqa-uc.a.run.app/v1/authorization
+    EVENTS_ENDPOINT=https://events-l3zyoxchqa-uc.a.run.app
```
mukobi: `EVENTS_ENDPOINT` matches the event**s** service, but I just noticed `USER_ENDPOINT` is missing the 's' from the user**s** service. I'm not too sure how big a change that might be to fix this, but maybe it should go in this PR or separate one depending.

cmei4444: Changed to USERS_ENDPOINT for consistency in 0682de4

pageserve/app.py
cmei4444: Thanks for the catch! Fixed in 0682de4

pageserve/app.py
cmei4444: Added in 5dfe2be

pageserve/app.py
```diff
@@ -76,34 +76,34 @@ def parsed_posts(posts):
...
+    # unreadably currently
+    return events_dict['events']
 
 
 def get_users_url():
     """Retrieves users URL, throws error if not found."""
-    url = os.environ.get("USER_ENDPOINT")
+    url = os.environ.get("USERS_ENDPOINT")
     if url is None:
         raise Exception("Users endpoint was undefined.")
```
mukobi: You might look into making this a more specific exception, see https://docs.python.org/3/library/exceptions.html. Maybe `NameError`?

cmei4444: Done in a9e2f33

pageserve/app.py
```diff
@@ -76,36 +76,36 @@ def parsed_posts(posts):
...
-               'event_time': '7-12-2019',
-               }]
-    return parsed_events(events)
-
-
-def parsed_events(events):
-    # TODO(cmei4444): implement parsing on events pulled from events service in
-    # a format for web display
-    return events
+    url = os.environ.get("EVENTS_ENDPOINT")
```
whaught: I tend to like making these 'constants' even when that's not totally a thing in python, but collecting these types of environment variable collection into one place might be nice

pageserve/app.py
```diff
@@ -76,36 +76,36 @@ def parsed_posts(posts):
...
+    r = requests.get(url, params={})
+    if r.status_code == 200:
+        return parsed_events(r.json())
+    else:
+        # TODO(cmei4444): handle error in a way that doesn't break page display
+        return "Error in getting events"
+
+
+def parsed_events(events_dict):
```
mayaspivak: Name this parse_events or get_parsed_events so that its an active description of what the function does.

pageserve/app.py
cmei4444: Would this be a good thing to do using app.config (similarly to app.config["COLLECTION"] in #27)?

mukobi: I think so. You could use `app.config["EVENTS_ENDPOINT"]` or similar.

pageserve/app.py
```diff
@@ -21,12 +21,20 @@
 
 app = Flask(__name__)
 
+ENDPOINTS = ['USERS_ENDPOINT', 'EVENTS_ENDPOINT']
+for ENDPOINT in ENDPOINTS:
+    if ENDPOINT in os.environ:
+        app.config[ENDPOINT] = os.environ.get(ENDPOINT)
+    else:
+        raise NameError("Endpoint {} not defined.".format(ENDPOINT))
```
cmei4444: Is this the best way to handle the env variables not being defined? Should this be a function instead? There will be at least one other endpoint.

pageserve/app.py
```diff
@@ -68,45 +77,37 @@ def get_posts():
     return parsed_posts(posts)
```
mukobi: Change to `parse_posts`?

pageserve/app.py
```diff
@@ -68,45 +77,37 @@ def get_posts():
...
-def get_users_url():
-    """Retrieves users URL, throws error if not found."""
-    url = os.environ.get("USER_ENDPOINT")
-    if url is None:
-        raise Exception("Users endpoint was undefined.")
-    return url
+    url = app.config['EVENTS_ENDPOINT']
+    r = requests.get(url, params={})
+    if r.status_code == 200:
+        return parsed_events(r.json())
```
mukobi: Change to `parse_events`?

pageserve/app.py
mukobi: I think it would look a little cleaner (and have fewer globals) if you refactored into a function like:
```python
app = Flask(__name__)

config_endpoints(endpoints=['USERS_ENDPOINT', 'EVENTS_ENDPOINT'])

def config_endpoints(endpoints):
    ...
```

pageserve/app.py
cmei4444: Thanks for catching, fixed in 6f44dde

pageserve/app.py
cmei4444: Thanks for catching, fixed in 6f44dde

pageserve/app.py
cmei4444: Done in 377b8cf

pageserve/app.py
cmei4444: Done in 377b8cf

pageserve/app.py
cmei4444: Done in f84f9a0

PR #27, opened 2019-07-24
(M) Test routes improve coverage

mukobi:
Untested code is broken code.

Add unit tests for routes in users and posts services and fix a few issues that were noticed because of it.

### DIFF
--- a/posts/app.py
+++ b/posts/app.py
@@ -50,7 +50,7 @@ def upload_new_post():
             "text":  request.form["text"],
             "files": [file for file in request.files.values()]
         }
-        return str(upload_new_post_to_db(post, POSTS_COLLECTION)), 201
+        return str(upload_new_post_to_db(post, app.config["COLLECTION"])), 201
     except BadRequestKeyError as error:
         return f"Invalid request. Required data: {REQUIRED_ATTRIBUTES}.", 400
     except AttributeError as error:
@@ -63,7 +63,7 @@ def upload_new_post():
 def get_all_posts():
     """Get all posts for the whole event."""
     # serialize otherwise nonserializable ObjectIDs
-    post_list = find_posts_in_db(POSTS_COLLECTION)
+    post_list = find_posts_in_db(app.config["COLLECTION"])
     return serialize_posts_to_json(post_list)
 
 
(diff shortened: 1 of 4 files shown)


posts/app.py
```diff
@@ -36,11 +36,11 @@
 def upload_new_post():
     """Make a new post upload to the server.
 
-    Post request body should contain:
+    Post request body should contain multipart/form-data with:
     event_id: id of the event to post to
     author_id: user id of the user making the post
     text: text to be sent
-    Post request files should contain all the files the user wants to upload
+    All the files the user wants to upload
```
cmei4444: Should these be in a list?

posts/app.py
```diff
@@ -227,7 +226,7 @@ def __getattribute__(self, _):
     return pymongo.MongoClient(mongodb_uri).posts_db.posts_collection
 
 
-POSTS_COLLECTION = connect_to_mongodb()  # None if can't connect
+app.config["COLLECTION"] = connect_to_mongodb()
```
cmei4444: Just wondering, was changing this for testing convenience or did the original way cause problems?

posts/app.py
mukobi: No, you actually are supposed to upload files in a request by sending a multipart/form-data content type (-F in curl) and sending each parameter individually. I.e. to send multiple files in curl, `curl -F foo=bar -F "file_1=/path/to/file_1" -F "file_2=/path/to/file_2" -X https://url.com`. It doesn't matter the names for the keys of the files you give, as files are automatically recognized and passed to Flask as `request.files`.

mukobi: I've updated the design doc to reflect this.

posts/app.py
mukobi: A bit of both. This makes it easier to test, and setting global configurations like the active database in Flask is preferred to use `app.config` (just a dict with a few special keys like `"TESTING"`) instead of global variables which are less clean.

users/test_routes.py
```diff
@@ -0,0 +1,90 @@
...
+    def test_missing_param(self):
+        """Try to get authorization but don't supply user_id."""
+        result = self.client.post("/v1/authorization")
+        self.assertEqual(result.status_code, 400)
+
+
+class TestAddUpdateUser(unittest.TestCase):
+    """Test add/update user endpoint PUT /v1/."""
+
+    def setUp(self):
```
cmei4444: Do you intend to add more tests for this endpoint?

mukobi: Yeah, the endpoint is just not implemented yet because I have to finish authentication first. Then, I'll add tests for it and should get users up to 100% coverage.

PR #28, opened 2019-07-25
(M) Sign in with google - front-end

mukobi:
Authentication requires 2 parts, a front-end UI/caller to set it up, and a back-end server to handle verifying tokens and storing user data in a db or cookies. This PR is just for the front-end and implements the "Google Sign In" button on all pages (using Jinja template inheritance). It won't work end-to-end yet because I need to implement the back-end server logic first.


pageserve/app.py
```diff
@@ -19,7 +19,12 @@
 
 from flask import Flask, render_template, request, Response
 
-app = Flask(__name__)
+app = Flask(__name__)  # pylint: disable=invalid-name
+
+app.config["GAUTH_CLIENT_ID"] = os.environ.get("GAUTH_CLIENT_ID")
+# TODO(mukobi) make env var $USERS_ENDPOINT stop at /v1 (remove /authorization)
+# Makes this cleaner after cmei444 merges PR #26
```
cmei4444: very minor note but my username has another 4 (cmei4444)

pageserve/templates/events.html
```diff
@@ -1,26 +1,39 @@
...
+<p>user does not have edit authorization</p>
+{% endif %}
+
+<hr>
+{% for event in events %}
+<p>Event name: {{event.name}}</p>
+<p>Event time: {{event.event_time}}</p>
+<p>Event description: {{event.description}}</p>
+<p>Created by {{event.author}} at {{event.created_at}}</p>
+<hr />
```
evankanderson: Why is this `<hr />` and not `<hr>`?

cmei4444: I think I originally added this, since they do the same thing we can just change to all `<hr>` for consistency

mukobi: I just copied that from what was there before, but `<hr>` should be sufficient. Done https://github.com/knative-portability/large-events/pull/28/commits/5bee11729e50e2b316084b3be2ed8024c539a788

pageserve/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/28/commits/d6d43f6fe46f4ba4a6dfe5007a7e881c5dd71280 https://github.com/knative-portability/large-events/pull/28/commits/9c624a1fc42c33797c886136fd50ba00133aade7. I make the env var `USERS_ENDPOINT` stop at /v1/ (you had it at /v1/authorization) so that it could more easily be used for other things by adding routes onto the url (e.g. add 'authenticate' for authentication).

PR #29, opened 2019-07-25
Increase coverage for events

cmei4444:
Added tests for the endpoint for retrieving all events.

Depends on previous changes from #26, most of the changes for this PR are in events/test_routes.py.


events/test_routes.py
```diff
@@ -0,0 +1,82 @@
...
+    def test_get_existing_events(self):
+        """Test retrieving all events from /v1/ endpoint."""
+        app.config["COLLECTION"].insert_many(FAKE_EVENTS)
+
+        response = self.client.get('/v1/')
+        self.assertEqual(response.status_code, 200)
+        data = json_util.loads(response.data)
+
+        self.assertEqual(len(data['events']), 2)
+        self.assertEqual(data['num_events'], 2)
```
mukobi: Consider changing hard coded `2` to `len(FAKE_EVENTS)`. Also, are you able to check that the values in `data['events']` and `FAKE_EVENTS` are equal, or is it returned differently?

events/test_routes.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def test_get_no_events(self):
+        """Test retrieving all events from /v1/ endpoint."""
```
mukobi: Change this docstring, it looks like it was copied from the previous test.

events/test_routes.py
```diff
@@ -0,0 +1,82 @@
...
+        self.client = app.test_client()
+
+    def test_post_event(self):
+        """Test posting of event."""
+        # TODO(cmei4444): test once retrieving form info is implemented
+
+
+class TestGetEventsRoute(unittest.TestCase):
+    """Test retrieve all events endpoint GET /v1/."""
+    def setUp(self):
```
mukobi: Docstring to explain why you need `setUp`?

events/test_routes.py
```diff
@@ -0,0 +1,82 @@
+import unittest
```
mukobi: Add module docstring and Apache 2.0 license.

cmei4444: I think a lot of my files are missing the license right now, I'll add it to all of them in another PR

events/test_routes.py
cmei4444: I've tried to check that the values are equal, but MongoDB tends to add some extra stuff that makes it hard to do - the returned events have timezones added onto the timestamps but `FAKE_EVENTS` doesn't get updated with it

events/test_routes.py
cmei4444: Added docstring in d8c42af

events/test_routes.py
cmei4444: Added docstring in d8c42af

events/test_routes.py
cmei4444: Updated docstring in d8c42af

events/test_routes.py
cmei4444: Changed to`len(FAKE_EVENTS)` in d8c42af

events/test_routes.py
```diff
@@ -0,0 +1,85 @@
...
+    'description': 'This event is missing an author!',
+    'event_time': EXAMPLE_TIME}
+INVALID_INFO_EXTRA_ATTRIBUTE = {
+    'name': 'invalid_event_extra',
+    'description': 'This event has an extra attribute.',
+    'author': 'admin',
+    'event_time': EXAMPLE_TIME,
+    'extra_attribute': 'I am invalid'}
+
+FAKE_INFO = [
```
evankanderson: Where is this used?

cmei4444: I was going to use it test adding events, I'll clarify why I need both fake info and fake events

cmei4444: Added comment in f73dfa0

events/test_routes.py
```diff
@@ -32,6 +32,7 @@
     'event_time': EXAMPLE_TIME,
     'extra_attribute': 'I am invalid'}
 
+# TODO(cmei4444): use for posting event information to /v1/add
```
evankanderson: My feedback was that you could introduce it when you need it for the tests; it's fine to leave this here not hooked up, but it's generally easier to not add these fixtures until you're actually writing the tests.

evankanderson:
This looks good, but encouraging you not to build test components you aren't using yet.

There's a term you may see referring to this: [YAGNI](https://www.google.com/search?q=define%3Ayagni)

events/test_routes.py
mukobi: For the equal values, you could maybe add an assertion function to the test class and use that to just test that a few key fields are the same while ignoring the other fields. Maybe something like
```python
class TestGetEventsRoute(unittest.TestCase):
    """Test retrieve all events endpoint GET /v1/."""
    def assert_events_are_equal(self, first, second):
        self.assertEqual(first['name'], second['name'])
        self.assertEqual(...)
        ...
```
And then call it like
```python
    def test_get_existing_events(self):
        ... get data and FAKE_EVENTS
        for i in range(len(data['events'])):
            self.assert_events_are_equal(data['events'][i], FAKE_EVENTS[i])
```

events/test_routes.py
cmei4444: I'll keep this in mind!

PR #31, opened 2019-07-25
(L) Gauth sign in users back end

mukobi:
Implement back-end authentication handling of Google login in users service.

### DIFF
--- a/users/app.py
+++ b/users/app.py
@@ -30,6 +30,9 @@ app = Flask(__name__)  # pylint: disable=invalid-name
 
 app.config["GAUTH_CLIENT_ID"] = os.environ.get("GAUTH_CLIENT_ID")
 
+VALID_GAUTH_TOKEN_ISSUERS = [
+    'accounts.google.com', 'https://accounts.google.com']
+
 
 @app.route('/v1/authorization', methods=['POST'])
 def get_authorization():
@@ -42,7 +45,7 @@ def get_authorization():
 
 
 @app.route('/v1/authenticate', methods=['POST'])
-def authenticate_user():
+def authenticate_and_get_user():
     """Authenticate user, upsert in the db, and return new user object.
 
     Request data:
@@ -90,8 +93,9 @@ def get_user_from_gauth_token(gauth_token):
     idinfo = id_token.verify_oauth2_token(
         gauth_token, requests.Request(), app.config["GAUTH_CLIENT_ID"])
 
-    if idinfo['iss'] not in ['accounts.google.com', 'https://accounts.google.com']:
-        raise ValueError('Wrong authentication token issuer.')
+    issuer = idinfo['iss']
+    if issuer not in VALID_GAUTH_TOKEN_ISSUERS:
+        raise ValueError(f'Invalid authentication token issuer "{issuer}".')
 
     # ID token is valid. Return the user object.
     return idinfo
--- a/users/app.py
+++ b/users/app.py
@@ -43,14 +43,17 @@ def add_update_user():
     gauth_token = request.form.get('gauth_token')
     if gauth_token is None:
         return "Error: You must supply a valid gauth_token.", 400
-    # TODO(mukobi) verify token and get the user object
     try:
-        user = None
-        if user is None:
-            raise NotImplementedError
+        user = get_user_from_gauth_token(gauth_token)
         return upsert_user_in_db(user, app.config["COLLECTION"]), 201
-    except NotImplementedError:
-        return "Error: Authentication not yet implemented", 400
+    except NotImplementedError as error:
+        return f"Error: {error}", 503
+
+
+def get_user_from_gauth_token(gauth_token):
+    """Validate the Google auth token and return it's user object."""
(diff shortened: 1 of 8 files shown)


users/test_authentication.py
```diff
@@ -0,0 +1,68 @@
...
+    of suppplying real tokens, we mock the functionality of
+    google.oauth2.id_token to simulate different responses from
+    different tokens.
+    """
+
+    def setUp(self):
+        """Patch google.oauth2.id_token."""
+        patcher = mock.patch('app.id_token')
+        self.verify_oauth2_token = patcher.start().verify_oauth2_token
+        self.addCleanup(patcher.stop)
```
cmei4444: Should this be `patcher.stop()` instead?

users/test_routes.py
```diff
@@ -76,14 +85,68 @@ def test_missing_param(self):
...
 
     def setUp(self):
         """Set up test client and seed mock DB for testing."""
         app.config["COLLECTION"] = mongomock.MongoClient().db.collection
         app.config["TESTING"] = True  # propagate exceptions to test client
         self.client = app.test_client()
+        # Patch google.oauth2.id_token
+        patcher = mock.patch('app.id_token')
+        self.verify_oauth2_token = patcher.start().verify_oauth2_token
+        self.addCleanup(patcher.stop)
```
cmei4444: Also should be `patcher.stop()`?

users/test_authentication.py
mukobi: No, that would immediately stop the patcher. `TestCase.addCleanup()` takes a function as a parameter and will call it automatically after `tearDown()`.

users/test_routes.py
mukobi: No, see above.

users/app.py
```diff
@@ -37,20 +41,60 @@ def get_authorization():
...
+        400: error message if authentication was not successful.
+    """
+    gauth_token = request.form.get('gauth_token')
+    if gauth_token is None:
+        return "Error: You must authenticate through Google.", 400
+    try:
+        idinfo = get_user_from_gauth_token(gauth_token)
+        user_object = {
+            "user_id": idinfo["sub"],
+            "name": idinfo["name"],
```
cmei4444: Maybe add a test where `"name"` isn't given - in the case where the user doesn't give the correct permissions it might not be returned.

mukobi: I'm not too sure how possible that is, since I think the smallest permission scope is 'profile' which contains an ID and a name, whereas things like email are additional scopes that aren't guaranteed. In any case, I added a test at https://github.com/knative-portability/large-events/pull/31/commits/62a4902f21f8190f09663d7d31b658d85cb2b807

users/app.py
```diff
@@ -37,20 +41,60 @@ def get_authorization():
...
-        return jsonify(error="You must supply a valid user in the body")
-    # TODO(mukobi) add or update the user in the database
-    added_new_user = True
-    # TODO(mukobi) get the new user object from the db to return
-    user_object = {
-        "user_id": "0", "username": "Dummy User", "edit_access": False
-    }
-    return jsonify(user_object), (201 if added_new_user else 200)
+@app.route('/v1/authenticate', methods=['POST'])
+def authenticate_user():
```
mayaspivak: maybe call this get_and_authenticate_user or authenticate_and_get_user

users/app.py
```diff
@@ -37,20 +41,60 @@ def get_authorization():
...
+            ...
+
+    Raises:
+        ValueError: Token is invalid.
+    """
+    # Authenticate token and match to client ID
+    idinfo = id_token.verify_oauth2_token(
+        gauth_token, requests.Request(), app.config["GAUTH_CLIENT_ID"])
+
+    if idinfo['iss'] not in ['accounts.google.com', 'https://accounts.google.com']:
```
mayaspivak: Make this a constant set at the top of the file.

users/app.py
```diff
@@ -37,20 +41,60 @@ def get_authorization():
...
+        raise ValueError('Wrong authentication token issuer.')
```
mayaspivak: Invalid instead of Wrong. Maybe return what the invalid value was? Is that safe to do? I don't know enough about authentication stuff.

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/31/commits/f2a8ccb636eee51d1e0c0a367c29b136ddf9486e

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/31/commits/f2a8ccb636eee51d1e0c0a367c29b136ddf9486e

users/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/31/commits/f2a8ccb636eee51d1e0c0a367c29b136ddf9486e. I think the token issuer is safe to return since it reflects the domain where the token was created and is simply checked against accounts.google.com

PR #32, opened 2019-07-26
Get event info from form data

cmei4444:
Added functionality for /v1/add endpoint to take in event info from POST request. 

Also added tests for the endpoint and fixed some formatting found along the way.


events/app.py
```diff
@@ -39,28 +40,24 @@ def __getattribute__(self, _):
...
-            response="Event info was entered incorrectly.",
-        )
+    except BadRequestKeyError:      # missing event attributes
+        return "Event info was entered incorrectly.", 400
     try:
         app.config["COLLECTION"].insert_one(event.dict)
-        return Response(
-            status=201,
-        )
+        return "Event added.", 201
```
mukobi: Why do you have this in a separate try block? It looks like if the first try block raises an exception, this try block will then execute but event could be undefined so you might get an uncaught exception. Maybe it would be better to move this into the first try block and list the excepts after.

events/test_routes.py
```diff
@@ -1,80 +1,89 @@
...
+
+VALID_DB_EVENT = {
     'name': 'valid_event',
     'description': 'This event is formatted correctly!',
     'author': 'admin',
-    'event_time': EXAMPLE_TIME}
-VALID_INFO_WITH_ID = {
+    'event_time': EXAMPLE_TIME,
+    'created_at': EXAMPLE_TIME,
+    'event_id': 'unique_event_id0'}
```
mukobi: Did we want to manually create event IDs or use the auto-generated IDs from MongoDB?

cmei4444: I think they should be defined by the DB but I'd have to change my Event class for that - will probably do it in a later PR

events/app.py
cmei4444: Done in c0a375b

events/app.py
```diff
@@ -39,28 +40,23 @@ def __getattribute__(self, _):
...
-    # TODO(cmei4444): Get event information from request data
-    info = request.form['info']
-    # TODO(cmei4444): Verify that user has event editing access
-    current_time = datetime.datetime.now()
-    info = build_event_info(info, current_time)
     try:
+        info = {
+            'name': request.form['event_name'],
+            'description': request.form['description'],
+            'author': request.form['author_id'],
```
mayaspivak: Is this input from the user? If so, we should use the id from the auth token instead.

events/test_routes.py
mukobi: Sounds good.

events/app.py
cmei4444: I was planning on retrieving the auth token ID in pageserve and posting it over with the rest of the information - not sure if this is the most secure way but I'm currently working on implementing it

mukobi: Agreed. That goes with the TODO on line 50. We'll need to get the gauth token from the request, make a request to the users service for authentication, check for authorization, then add the ID to info. Maybe that's better left to a separate PR.

cmei4444: It's probably best to do that in another PR, clarified the TODO in 0e29b99

mayaspivak: Sounds good.

PR #33, opened 2019-07-26
(M) User authenticate preserve is_organizer in db

mukobi:
Previously (in #31), the authentication endpoint would replace the value of the `is_organizer` field with a default of False in the database whenever a user signed in. This uses the special `$setOnInsert` Mongo command to only set this value on a new user creation. Otherwise, it preserves the existing value of `is_organizer` in the database for the user. This allows us to use the /v1/authenticate endpoint as a secure way not only to create new users when logging in with OAuth but also to query the exact `user_id`, `name`, or `is_organizer` fields of a user from another service using their Google Auth ID token without worrying about overriding their authorization level in the database.

This effectively makes the /v1/authorization endpoint (for getting the authentication of a user based on their `user_id`) almost useless, as /v1/authentication should be preferred since it's simpler, based on secure authorization tokens, and only lets the user see the authorization of themselves. The only case for keeping the /v1/authorization endpoint might be if we wanted to also show a list of all the event organizers somewhere, though that might do better in a separate endpoint that simply queries the db for all organizers.

### DIFF
--- a/users/test_user_upsertion.py
+++ b/users/test_user_upsertion.py
@@ -49,6 +49,8 @@ class TestUserUpsertion(unittest.TestCase):
         # original user attributes matches found user attributes
         self.assertEqual(user_to_insert["user_id"], found_user["user_id"])
         self.assertEqual(user_to_insert["name"], found_user["name"])
+        # authorization false by default
+        self.assertFalse(found_user["is_organizer"])
         # user returned from app.upsert_user_in_db matches found user exactly
         returned_user = self.mock_collection.find_one(upserted_id)
         self.assertEqual(returned_user, found_user)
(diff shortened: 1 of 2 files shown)


users/test_user_upsertion.py
```diff
@@ -50,12 +49,38 @@ def test_insert_valid_user(self):
         # original user attributes matches found user attributes
         self.assertEqual(user_to_insert["user_id"], found_user["user_id"])
         self.assertEqual(user_to_insert["name"], found_user["name"])
-        self.assertEqual(
-            user_to_insert["is_organizer"], found_user["is_organizer"])
```
cmei4444: To make sure the field is set correctly, you should probably still check that `found_user["is_organizer"]` is `False` here.

users/test_user_upsertion.py
```diff
@@ -110,8 +129,6 @@ def test_multiple_upserts_is_one_insert(self):
...
```
cmei4444: Also check that `found_user["is_organizer"]` is false here.

users/test_user_upsertion.py
mukobi: Done https://github.com/knative-portability/large-events/pull/33/commits/775384ce54f94cc2e9a0fb16c2c65b0c302de55a

users/test_user_upsertion.py
mukobi: Done https://github.com/knative-portability/large-events/pull/33/commits/775384ce54f94cc2e9a0fb16c2c65b0c302de55a

PR #34, opened 2019-07-26
Create front-end form entry UI

cmei4444:
Created UI for adding events and posts, along with endpoints in pageserve to handle both.

Currently working on testing routes for pageserve (for another PR), will add tests for these new routes too.


pageserve/templates/events.html
```diff
@@ -36,4 +36,28 @@ <h1>{% block title %}Sub-Events{% endblock title %}</h1>
...
+        <label for="description">Description:</label>
+        <textarea type="text" rows = "5" cols = "60" id="description" name="description">
+          Enter details here.
+        </textarea>
+    </div>
+    <div>
+        <label for="event_time">Event Time:</label>
+        <input type="datetime-local" id="event_time" name="event_time">
+    </div>
+    <input type="hidden" id="author" name="author_id" value="{{user}}">
```
mukobi: What might be better than a hidden field is not sending this from the front end and instead having `add_event()` handle this. In `add_event()`, you could make it get the user object (I'm working on writing a user authentication helper function to make this one call, for now you could just use a placeholder). Then if the returned user exists and is an organizer, you can send it along to the request with `data={**request.form, author_id=user['user_id']}`. Same goes for the post form, but you wouldn't have to check for `is_organizer` in `add_event()`.

pageserve/templates/events.html
```diff
@@ -36,4 +36,28 @@ <h1>{% block title %}Sub-Events{% endblock title %}</h1>
 {% endfor %}
 
 <a href='/v1/'>See all posts</a>
-{% endblock content%}
\ No newline at end of file
+
+<h2>Create New Event</h2>
```
mukobi: Do you think we might want event adding on a separate page from showing all the events?

pageserve/app.py
```diff
@@ -52,11 +52,34 @@ def index():
...
 
 
-@app.route('/v1/events')
+@app.route('/v1/add_post', methods=['POST'])
+def add_post():
+    """Add post by calling posts service."""
+    url = app.config['POSTS_ENDPOINT'] + 'add'
+    r = requests.post(url, data=request.form)
+    if r.status_code == 201:
+        return "Post successfully added"
```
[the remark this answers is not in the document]
evankanderson: You may find it easier to return the same "shape" of result on both success and failure (e.g. always a string, or always a JSON object).

You might want to document the expected parameters and return format in a docstring here.

pageserve/app.py
```diff
@@ -52,11 +52,34 @@ def index():
...
+    else:
+        return r.content
+
+
+@app.route('/v1/add_event', methods=['POST'])
+def add_event():
```
evankanderson: Does the ordering of this file make sense?

Right now, it looks like:

* config_endpoints
* GET /v1
* POST /v1/add_post
* POST /v1/add_event
* GET /v1/events
* get/parse posts
* get/parse events
* get_user_info

pageserve/templates/events.html
evankanderson: Single-page app seems reasonable, if the list extends as you add a new event.

pageserve/templates/events.html
evankanderson: +1 for checking this with authentication server-side.

pageserve/app.py
mukobi: A few of our other files also have quite poor ordering. I think it might make sense to standardize something like
- Routes alphabetized by endpoint then method
- Helper functions alphabetized
- Configuration functions (config endpoints, connect to db) alphabetized
- if __name__ == "__main__"

pageserve/app.py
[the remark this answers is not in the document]
cmei4444: Clarified in docstrings in 726a5e5

pageserve/app.py
```diff
@@ -52,11 +52,34 @@ def index():
...
-@app.route('/v1/events')
+@app.route('/v1/add_post', methods=['POST'])
+def add_post():
+    """Add post by calling posts service."""
+    url = app.config['POSTS_ENDPOINT'] + 'add'
+    r = requests.post(url, data=request.form)
+    if r.status_code == 201:
+        return "Post successfully added"
+    else:
+        return r.content
```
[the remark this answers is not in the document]
cmei4444: Done in 3b470a5

pageserve/templates/events.html
cmei4444: Good idea, done in 3b470a5

pageserve/templates/events.html
cmei4444: I'm going to leave it as single-page for now

pageserve/app.py
cmei4444: Alphabetizing might not make the most logical ordering of functions, depending on naming. For now I'll re-order the endpoints and move the config functions.

Wondering if it's worth another PR to re-order files?

pageserve/app.py
```diff
@@ -68,10 +51,55 @@ def show_events():
...
+    Received form data should be multipart/form-data and contain:
+        event_id: id of the event to post to
+        text: text content of post
+        All files to be uploaded (can be multiple)
+
+    Response:
+        string response and response code returned by posts service
+    """
+    url = app.config['POSTS_ENDPOINT'] + 'add'
+    # TODO: send authenticated user instead of placeholder
```
evankanderson: This TODO doesn't seem needed here with the get_user function?

evankanderson: (get_user may end up needing the request, though)

pageserve/app.py
```diff
@@ -68,10 +51,55 @@ def show_events():
...
+    Received form data should contain:
+        event_name: name of the created event
+        description: description of event
+        event_time: time of event
+
+    Response:
+        string response and response code returned by events service
+    """
+    url = app.config['EVENTS_ENDPOINT'] + 'add'
+    # TODO: send authenticated user instead of placeholder
```
evankanderson: ditto on this TODO

pageserve/app.py
evankanderson: This sort of refactor is often good as a fast-followup in a separate PR when it moves existing functions (so that the reviewer can easily distinguish between changed/added code and simple cleanups).

The amount of change in this PR isn't a big deal, but if you have (for example) a Java class with 20 methods and you reorganize them while adding 2 more, it can get tricky to review.

PR #36, opened 2019-07-29
Coverage for GET endpoints of pageserve

cmei4444:
Added test coverage for the GET endpoints in pageserve, which return page templates. 

Installs packages and sets up testing for Flask render_templates.


pageserve/test_routes.py
```diff
@@ -0,0 +1,68 @@
...
+        return app
+
+    def setUp(self):
+        """Set up test client."""
+        self.client = app.app.test_client()
+
+    @patch('app.get_user', return_value=EXAMPLE_USER)
+    @patch('app.has_edit_access', return_value=True)
+    @patch('app.get_posts', return_value=EXAMPLE_POSTS)
+    def test_index(self, mock_posts, mock_access, mock_user):
```
cmei4444: Not really sure if this is the best way to patch all these functions for testing, would like some input.

pageserve/requirements.txt
```diff
@@ -1,2 +1,4 @@
 flask
-requests
\ No newline at end of file
+requests
+Flask-Testing
+blinker
```
mukobi: What's blinker for? Is it something Flask-Testing depends on? If so and if you need to explicitly add it to your requirements, it might be good to add a comment.

pageserve/test_routes.py
```diff
@@ -0,0 +1,68 @@
...
+from flask import Flask
+from flask_testing import TestCase
+import app
+
+EXAMPLE_USER = "app_user"
+EXAMPLE_POSTS = ['example', 'posts', 'list']
+EXAMPLE_EVENTS = ['example', 'events', 'list']
+
+
+class TestRoutes(TestCase):
```
mukobi: Have you thought of splitting the different endpoints into their own classes? I think that makes files like this (which will end up with a whole lot of tests) cleaner to navigate. Or if you think there aren't enough tests for each of these endpoints to have their own test class, alternatively you could make this `class TestGETRoutes` or similar.

pageserve/test_routes.py
mukobi: I don't like those unused parameters, and pylint should get angry at you too if you run it. Instead, does replacing those with single underscores work? Alternatively, in your body you could also nest patches as context managers:
```python
with patch('app.get_user', return_value=EXAMPLE_USER):
    with patch('app.has_edit_access', return_value=True):
        with patch('app.get_posts', return_value=EXAMPLE_POSTS):
            ...
```
but that just looks ugly.

cmei4444: If the concern is pylint, then replacing them with each with single underscores is worse since they're duplicates. Prefixing them with an underscore gets pylint to ignore them, but that doesn't really address the issue of the unused parameters.

mukobi: What if you specify in the patches to replace it with a magic mock with each specific return value. That might allow you to remove the parameters from the function since you are explicitly created the replacement mock objects in place.
```suggestion
    from unittest.mock import MagicMock
    
    @patch('app.get_user', MagicMock(return_value=EXAMPLE_USER))
    @patch('app.has_edit_access', MagicMock(return_value=True))
    @patch('app.get_posts', MagicMock(return_value=EXAMPLE_POSTS))
    def test_index(self):
```

cmei4444: That lets me get rid of the parameters, thanks

pageserve/requirements.txt
cmei4444: Flask-Testing depends on it, added comment in 1bf1c5d

pageserve/test_routes.py
cmei4444: Renamed in 1bf1c5d

pageserve/test_routes.py
```diff
@@ -0,0 +1,70 @@
...
+    @patch('app.get_user', MagicMock(return_value=EXAMPLE_USER))
+    @patch('app.has_edit_access', MagicMock(return_value=True))
+    @patch('app.get_posts', MagicMock(return_value=EXAMPLE_POSTS))
+    def test_index(self):
+        """Checks index page is rendered correctly by GET /v1/."""
+        response = self.client.get('/v1/')
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed('index.html')
+
+        self.assertContext('user', EXAMPLE_USER)
```
evankanderson: This is cool -- I'm assuming it's part of `Flask-Testing`?

cmei4444: Yeah, it seems to be the only good way to test that the templates receive the right info.

PR #38, opened 2019-07-30
Tests for events DB error and minor cleanup

cmei4444:
Added tests for the case of an undefined events DB connection to improve coverage.

Also did some minor cleanup of events, also mainly to improve coverage.

### DIFF
--- a/events/app.py
+++ b/events/app.py
@@ -68,7 +68,7 @@ def build_event_info(info, time):
 @app.route('/v1/edit/<event_id>', methods=['PUT'])
 def edit_event(event_id):
     """Edit the event with the given id."""
-    pass    # pragma: no cover
+    pass
 
 
 @app.route('/v1/', methods=['GET'])
@@ -97,13 +97,13 @@ def build_events_dict(events):
 @app.route('/v1/search', methods=['GET'])
 def search_event():
     """Search for the event with the given name in the DB."""
-    pass    # pragma: no cover
+    pass
 
 
 @app.route('/v1/<event_id>', methods=['PUT'])
 def get_one_event(event_id):
     """Retrieve one event by event_id."""
-    pass    # pragma: no cover
+    pass
 
 
 if __name__ == "__main__":  # pragma: no cover
--- a/events/app.py
+++ b/events/app.py
@@ -4,7 +4,7 @@ import json
 import pymongo
 from bson import json_util
 
-from flask import Flask, request, Response
+from flask import Flask, request
 from werkzeug.exceptions import BadRequestKeyError
 from eventclass import Event
 
@@ -17,7 +17,6 @@ class DBNotConnectedError(ConnectionError):
 
 def connect_to_mongodb():   # pragma: no cover
     # TODO(cmei4444): restructure to be consistent with other services
-    # TODO(cmei4444): test with deployed service
     """Connects to MongoDB Atlas database.
 
     Returns events collection if connection is successful, and None otherwise.
@@ -78,8 +77,7 @@ def get_all_events():
         events = app.config["COLLECTION"].find({})
         events = [Event(**ev).dict for ev in events]
         events_dict = build_events_dict(events)
-        # handle objects from MongoDB (e.g. ObjectID) that aren't JSON
-        # serializable
+        # handle MongoDB objects (e.g. ObjectID) that aren't JSON serializable
         return json.loads(json_util.dumps(events_dict))
     except DBNotConnectedError as e:
         return "Events database was undefined.", 500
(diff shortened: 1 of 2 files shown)


events/test_routes.py
```diff
@@ -59,6 +60,13 @@ def test_add_invalid_event(self):
 
         self.assertEqual(self.coll.count_documents({}), 0)
 
+    def test_db_not_defined(self):
```
mukobi: Docstring? It could help to run `pylint -v app` and fix any style errors before requesting a PR.

events/test_routes.py
```diff
@@ -93,3 +101,8 @@ def test_get_no_events(self):
 
         self.assertEqual(len(data['events']), 0)
         self.assertEqual(data['num_events'], 0)
+
+    def test_db_not_defined(self):
```
mukobi: Docstring?

events/test_routes.py
cmei4444: Added 6775c69

events/test_routes.py
cmei4444: Added 6775c69

events/app.py
```diff
@@ -69,7 +68,6 @@ def build_event_info(info, time):
 @app.route('/v1/edit/<event_id>', methods=['PUT'])
 def edit_event(event_id):
     """Edit the event with the given id."""
-    pass
```
evankanderson: This is a handy hint that the code isn't implemented yet; I would leave it.

events/app.py
```diff
@@ -99,14 +96,12 @@ def build_events_dict(events):
 @app.route('/v1/search', methods=['GET'])
 def search_event():
     """Search for the event with the given name in the DB."""
-    pass
```
evankanderson: Ditto here

events/test_routes.py
```diff
@@ -59,6 +60,14 @@ def test_add_invalid_event(self):
 
         self.assertEqual(self.coll.count_documents({}), 0)
 
+    def test_db_not_defined(self):
+        """Test adding event when DB connection is undefined."""
+        self.coll.insert_one = MagicMock(side_effect=DBNotConnectedError)
+        response = self.client.post('/v1/add', data=VALID_REQUEST_INFO)
```
evankanderson: Nice test!

One other way to do this would be to patch the mongo URL to choose an invalid connection string: https://github.com/mongomock/mongomock/blob/develop/tests/test__mongomock.py#L54

But that might require restructuring your code.

mukobi: If you want to try this, maybe something like this would work.
```python
del os.environ["MONGODB_URI"]  # unset db config variable
app.config["COLLECTION"] = connect_to_mongodb()  # will throw DBNotConnectedError on access
response = self.client.post('/v1/add', data=VALID_REQUEST_INFO)  # will throw inside handler
```

events/app.py
cmei4444: Added back in ea7f1b5

events/app.py
cmei4444: Added back in ea7f1b5

events/test_routes.py
cmei4444: Seems to work, thanks! I just need to make sure to replace the MONGODB_URI variable afterwards otherwise later tests will fail.

cmei4444: Nevermind, seems to work when testing on Travis - it doesn't seem to want me to mess with env variables. I think I'll just stick to the previous method.

evankanderson: Here's an example of using a context manager to set environment variables (use the last comment in the gist): https://gist.github.com/igniteflow/7267431

events/app.py
```diff
@@ -69,7 +68,7 @@ def build_event_info(info, time):
 @app.route('/v1/edit/<event_id>', methods=['PUT'])
 def edit_event(event_id):
     """Edit the event with the given id."""
-    pass
+    pass    # pragma: no cover
```
evankanderson: It's fine to leave these as tested by coverage -- it's a sign that something isn't entirely working. 😁

events/test_routes.py
mukobi: The error in Travis was because `del os.environ["MONGODB_URI"]` fails if `"MONGODB_URI"` is not a key in the environment. Either way you choose to do it, you might have to check if `os.environ["MONGODB_URI"]` is set first.

evankanderson: I think the `contextlib` approach should be able to work around this:

```from contextlib import contextmanager

@contextmanager
def environ(env):
    """Temporarily set environment variables inside the context manager and
    fully restore previous environment afterwards
    """
    original_env = {key: os.getenv(key) for key in env}
    os.environ.update(env)
    try:
        yield
    finally:
        for key, value in original_env.items():
            if value is None:
                del os.environ[key]  # You could set the value to "WHAT, see line XYZ" instead of del
            else:
                os.environ[key] = value
```

cmei4444: Implemented with context manager in 6543e0e (with fix in 64c409b)

PR #41, opened 2019-07-31
(M) Users add edit authorization route

mukobi:
Add and test route to change authorization level of users from users service

### DIFF
--- a/users/test_routes.py
+++ b/users/test_routes.py
@@ -46,19 +46,24 @@ IDINFO_MISSING_NAME = {
     "sub": "I have an ID but no name"}
 DUMMY_GAUTH_REQUEST_DATA = {"gauth_token": "fake_token_0123"}
 
-UPDATER_ID = "authorized-self-id"
+AUTHORIZED_UPDATER_ID = "authorized-self-id"
 AUTHORIZED_UPDATER = {
-    "user_id": UPDATER_ID,
-    "name": UPDATER_ID,
+    "user_id": AUTHORIZED_UPDATER_ID,
+    "name": AUTHORIZED_UPDATER_ID,
     "is_organizer": True}
+NON_AUTHORIZED_UPDATER_ID = "non-authorized-self-id"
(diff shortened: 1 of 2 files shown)


users/test_routes.py
```diff
@@ -88,14 +97,131 @@ def test_missing_param(self):
...
+        self.assertTrue(app.find_authorization_in_db(
+            target_user_id, app.app.config["COLLECTION"]))
+        self.client.post("/v1/authorization/update", data={
+            "target_user_id": target_user_id,
+            "is_organizer": False,
+            **DUMMY_GAUTH_REQUEST_DATA})
+        self.assertFalse(app.find_authorization_in_db(
+            target_user_id, app.app.config["COLLECTION"]))
+
+    def test_not_authorized(self):
```
cmei4444: Maybe split this into separate tests? And create an unauthorized updater user so you don't have the change the updater's authorization to False.

mukobi: Done https://github.com/knative-portability/large-events/pull/41/commits/0e17603bf64465058b50e9767acb7f48bfff6235

users/test_routes.py
```diff
@@ -47,10 +47,14 @@
 DUMMY_GAUTH_REQUEST_DATA = {"gauth_token": "fake_token_0123"}
 
 UPDATER_ID = "authorized-self-id"
-UPDATER_OBJECT = {
+AUTHORIZED_UPDATER = {
     "user_id": UPDATER_ID,
     "name": UPDATER_ID,
     "is_organizer": True}
+NON_AUTHORIZED_UPDATER = {
+    "user_id": UPDATER_ID,
```
cmei4444: `UPDATER_ID` is `"authorized-self-id"` right now - maybe create a second ID as well and add both updaters to the DB - would simplify your tests as well because you don't have to update the updater user

mukobi: Done https://github.com/knative-portability/large-events/pull/41/commits/155b209516d739a785b9810fb7b136833c0d3d9e

PR #43, opened 2019-07-31
Display posts

cmei4444:
Get posts from posts service and display content on index.html template.

Also added tests for getting posts/events from their respective services and error handling if they can't be retrieved, displaying an error message instead of the template.


pageserve/app.py
```diff
@@ -29,14 +29,17 @@ def index():
...
-    posts = get_posts()
-    return render_template(
-        'index.html',
-        posts=posts,
-        auth=is_auth,
-        user=user,
-        app_config=app.config
-    )
+    try:
+        posts = get_posts()
```
mukobi: This line and line 53 can be inline instead of a temp variable

pageserve/app.py
```diff
@@ -29,14 +29,17 @@ def index():
...
+        return render_template(
+            'index.html',
+            posts=posts,
+            auth=is_auth,
+            user=user,
+            app_config=app.config
+        )
+    except RuntimeError:
+        return "Unable to get posts from posts service.", 500
```
mukobi: Since you are giving a message to your RuntimeError on line 117, maybe it would be better to just pass the error along instead of redefining a new error message. This way, both `except`s in `index()` and `show_events()` would look the same.
```python
except RuntimeError as error:
    return str(error), 500
```

pageserve/app.py
```diff
@@ -104,28 +110,27 @@ def add_event():
...
+    Args:
+        events_dict: JSON returned by posts service, includes:
+            posts (list): list of posts
+            num_posts (int): number of posts returned
+
+    Returns:
+        list: parsed list of posts.
+    """
+    # TODO(cmei4444): implement parsing on posts - timestamps are formatted
+    # unreadably currently
```
mukobi: It sounds like you've done more research into how `datetime`s work in the db than I have, what kind of parsing would be needed? Depending on what we want to do, it might be simple to do the parsing in posts and events before we upload stuff to the database and just store them as strings. If you try out `str(datetime.datetime.now())` it actually returns a fairly readable time string.

cmei4444: I think I actually figured it out - the events service at least is giving Unix times in milliseconds, which are pretty easy to parse: 
```
ts = int(unix_time)/1000
event['created_at'] = datetime.utcfromtimestamp(ts).strftime('%Y-%m-%d %H:%M:%S')
```

In the DB datetimes are actually stored pretty readably (from looking at the MongoDB atlas site), the issue happens when you try to JSON serialize them. 
Storing them as strings would be easier for testing at this point though so it honestly might be preferable.

mukobi: I was just testing things and `datetime.datetime.utcnow().isoformat(" ", "seconds")` returns `'2019-07-31 23:52:18'`. I highly recommend we use this when creating timestamps and store the times as strings like that in the DB. If we don't care about globalization, we can drop the `utcnow()` to just `now()` for local time, but I figure it's probably good principle for a social app like this to use standard time.

cmei4444: That seems good, let's go with that. I'll implement that for events in another PR.

mukobi: Cool, I'll implement it for posts in another PR.

PR #44, opened 2019-07-31
Reorder events app.py functions

cmei4444:
This is really more to discuss file ordering - I used this order here, imitating the pageserve app.py order:

GET endpoints
POST endpoints
PUT endpoints
helper functions
config functions
__main__

Within those categories I don't feel super strongly about a way to order them as long as it seems logical, would like input for this file and other app.py files as well.

### DIFF
--- a/events/app.py
+++ b/events/app.py
@@ -11,30 +11,23 @@ from eventclass import Event
 app = Flask(__name__)
 
 
-class DBNotConnectedError(ConnectionError):
-    """Raised when not able to connect to the db."""
+@app.route('/v1/', methods=['GET'])
+def get_all_events():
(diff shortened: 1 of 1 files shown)


cmei4444:
> LGTM. I agree with the high-level ordering you chose here. Inside each of those categories, I think we should alphabetize by method name because it doesn't really matter too much and just makes things a little easier to find. From my quick look, it appears that you alphabetized within each category on this event/app.py. I think reordering is definitely one of the cleanup things we should do before we release this, but I also think you should wait on more reordering PRs until after some of the other bigger PRs touching the same files get merged to avoid painful merge conflicts.

I agree with your last point, I'm definitely not going to touch pageserve for now since we're both working on it.

PR #45, opened 2019-08-01
(XS) Posts respresent timestamp as string

mukobi:
Upload posts timestamps to db as UTC time strings with second accuracy.

### DIFF
--- a/posts/app.py
+++ b/posts/app.py
@@ -169,7 +169,7 @@ def generate_timestamp():
     Returns:
         datetime: the current time.
     """
-    return datetime.datetime.utcnow().isoformat(" ", "seconds")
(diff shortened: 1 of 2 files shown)


posts/app.py
```diff
@@ -169,7 +169,7 @@ def generate_timestamp():
     Returns:
         datetime: the current time.
     """
-    return datetime.datetime.now()
+    return datetime.datetime.utcnow().isoformat(" ", "seconds")
```
whaught: maybe a named parameter for  delimiter=" " or whatever this means

mukobi: Done https://github.com/knative-portability/large-events/pull/45/commits/a88de4b2953206978ee4c6a11b853ea9f29b65ed

PR #46, opened 2019-08-01
Events creation time timestamp as string

cmei4444:
Similar to #45, changes the format of the created_at time for events to a string. Modified comparison of times for equality, leaving the previous functionality comparing datetime objects. Also added some tests.

This PR depends on #44, all relevant changes for this one are in the last commit.

### DIFF
--- a/events/app.py
+++ b/events/app.py
@@ -42,7 +42,8 @@ def add_event():
         }
         # TODO(cmei4444): Athenticate user and verify that user has event
         # editing access
-        current_time = datetime.datetime.utcnow().isoformat(" ", "seconds")
+        current_time = datetime.datetime.utcnow().isoformat(sep=" ",
+                                                            timespec="seconds")
         info = build_event_info(info, current_time)
         event = Event(**info)
         app.config["COLLECTION"].insert_one(event.dict)
--- a/events/app.py
+++ b/events/app.py
@@ -42,7 +42,7 @@ def add_event():
         }
         # TODO(cmei4444): Athenticate user and verify that user has event
         # editing access
-        current_time = datetime.datetime.now()
+        current_time = datetime.datetime.utcnow().isoformat(" ", "seconds")
         info = build_event_info(info, current_time)
         event = Event(**info)
         app.config["COLLECTION"].insert_one(event.dict)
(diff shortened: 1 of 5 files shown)


events/app.py
```diff
@@ -49,7 +42,7 @@ def add_event():
         }
         # TODO(cmei4444): Athenticate user and verify that user has event
         # editing access
-        current_time = datetime.datetime.now()
+        current_time = datetime.datetime.utcnow().isoformat(" ", "seconds")
```
whaught: same comment about parameter labels on this call. That " " is hard to read/understand what it means.

cmei4444: Clarified in 5fdd6a3

events/eventclass.py
```diff
@@ -15,10 +16,17 @@ def equal_times(time_1, time_2):
...
-    time_2 = time_2.replace(microsecond=(time_2.microsecond // 1000) * 1000)
-    return time_1 == time_2
+    if type(time_1) is type(time_2):
+        if isinstance(time_1, datetime.datetime):
+            time_1 = time_1.replace(
+                microsecond=(time_1.microsecond // 1000) * 1000)
+            time_2 = time_2.replace(
+                microsecond=(time_2.microsecond // 1000) * 1000)
+        return time_1 == time_2
+    return False
```
mukobi: If we're doing time as strings, why do you need this replacing still? Are you planning on also supporting datetimes at the same time?

events/test_routes.py
```diff
@@ -9,7 +9,7 @@
 from app import app, os, connect_to_mongodb
 
 EXAMPLE_TIME = datetime.datetime(
-    2019, 6, 11, 10, 33, 1, 100000).isoformat(" ", "seconds")
+    2019, 6, 11, 10, 33, 1, 100000).isoformat(sep=" ", timespec="seconds")
```
mukobi: You might want to make this `EXAMPLE_TIME_STRING` since it's a string so it matches the global var naming in test_events_class.py.

events/eventclass.py
cmei4444: I guess I wanted to leave functionality there in case times had to be compared as datetimes in the backend, but I don't think it's used anywhere right now so I can remove it.

events/test_routes.py
cmei4444: Done for all test files in df59c2b

PR #47, opened 2019-08-01
(S) Query event by name

cmei4444:
Added ability to query events by name.

### DIFF
--- a/events/app.py
+++ b/events/app.py
@@ -33,7 +33,7 @@ def search_event():
 
         events = [Event(**ev).dict for ev in events]
         events_dict = build_events_dict(events)
-        # handle MongoDB objects (e.g. ObjectID) that aren't JSON serializable
+        # handles MongoDB objects (e.g. ObjectID) that aren't JSON serializable
         return json.loads(json_util.dumps(events_dict))
     except BadRequestKeyError:      # missing event attributes
         return "Event name was entered incorrectly.", 400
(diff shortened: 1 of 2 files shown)


events/app.py
```diff
@@ -27,7 +27,18 @@ def get_all_events():
 @app.route('/v1/search', methods=['GET'])
 def search_event():
     """Search for the event with the given name in the DB."""
-    pass
+    try:
+        event_name = request.args['name']
+        events = app.config["COLLECTION"].find({'name': event_name})
+
+        events = [Event(**ev).dict for ev in events]
+        events_dict = build_events_dict(events)
```
evankanderson: Why can't this just take Event objects, rather than Event objects rendered as dictionaries?

events/app.py
```diff
@@ -27,7 +27,18 @@ def get_all_events():
...
+        # handle MongoDB objects (e.g. ObjectID) that aren't JSON serializable
```
evankanderson: Is this a TODO, or an explanation of what's going on?

events/app.py
evankanderson: In particular, you have the same dictionary comprehension on line 19 as here.

events/app.py
```diff
@@ -82,12 +79,12 @@ def build_event_info(info, time):
     return {**info, 'created_at': time}
 
 
-def build_events_dict(events):
+def build_events_dict(events_cursor):
```
evankanderson: Good call to name this a cursor to make it clear that the call may end up moving the underlying cursor.

PR #48, opened 2019-08-01
(XS) Better coverage reporting

mukobi:
Use branch coverage reporting (https://coverage.readthedocs.io/en/v4.5.x/branch.html) and run coverage on all python files (including test files).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_script:
   - pip install -r requirements.txt
 script:
   # run coverage on files testing the app module
-  - coverage run --branch --omit /home/* -m unittest discover
+  - coverage run --branch --omit */home/* -m unittest discover
   - cp .coverage* ../
 
 after_success:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -32,13 +32,13 @@ before_script:
   - pip install -r requirements.txt
 script:
   # run coverage on files testing the app module
-  - coverage run --source app -m unittest discover
-  - cp .coverage* ../
+  - coverage run --branch --omit="*/virtualenv/*","*.html" -m unittest discover
```
cmei4444: I think coverage automatically ignores non-python files, so you don't need to ignore html files.

mukobi: I tried that, but it broke on Pageserve when running Coveralls. It tried to run the HTML then broke on line 1 saying it wasn't Python. Pageserve still passed the Travis build for some reason, but it wasn't able to report coverage to Coveralls. Really annoying, but this fixes it.

cmei4444: I haven't been able to find out why that happens, so that seems like the best way to handle it.

PR #49, opened 2019-08-01
(S) Get event by ID

cmei4444:
Added ability to retrieve one event by its unique event ID. 

Slightly restructured the way events are converted into dict form so an `event_id` field doesn't get stored in MongoDB, which could have led to duplicate or conflicting info. The `get_dict` function now just moves the info in `event_id` to a field called `_id`.

Namedtuple doesn't allow `_id` as a field so I couldn't just modify the required attributes.

### DIFF
--- a/events/app.py
+++ b/events/app.py
@@ -62,7 +62,14 @@ def edit_event(event_id):
 @app.route('/v1/<event_id>', methods=['PUT'])
 def get_one_event(event_id):
     """Retrieve one event by event_id."""
-    pass
+    try:
+        events = app.config["COLLECTION"].find({'_id': event_id})
+        events = [Event(**ev).dict for ev in events]
+        events_dict = build_events_dict(events)
+        # handle MongoDB objects (e.g. ObjectID) that aren't JSON serializable
+        return json.loads(json_util.dumps(events_dict))
+    except DBNotConnectedError as e:
+        return "Events database was undefined.", 500
 
 
 def build_event_info(info, time):
--- a/events/app.py
+++ b/events/app.py
@@ -2,7 +2,7 @@ import os
 import datetime
 import json
 import pymongo
-from bson import json_util
+from bson import json_util, ObjectId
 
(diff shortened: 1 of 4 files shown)


cmei4444:
> I think there's a problem with your post adding logic. MongoDB defaults unique `ObjectID`s as the type for the `_id` field, but it allows you to override that to be whatever datatype you want as long as it is unique and not an array. This can sometimes be useful, but I think you probably want all event IDs to be unique randomly generated `ObjectID`s. To do this, your uploading function must construct an `ObjectID` from the string parameter that was passed to the request (see [here](https://github.com/knative-portability/large-events/blob/master/posts/app.py)), and your tests need to use valid string representations of `ObjectID`s which are 24-char hex strings (see line 189 [here](https://github.com/knative-portability/large-events/blob/master/posts/test_routes.py)).

Do you mean I should create an ObjectID from the string passed to get_one_event? That seems the most similar to what you did in the posts app.py file.

mukobi:
> Do you mean I should create an ObjectID from the string passed to get_one_event? That seems the most similar to what you did in the posts app.py file.

Yeah, but then you'll find your tests won't work if you are testing with fake ObjectIDs unless they are formatted as 24-char hex strings. Alternatively, I guess you could ignore the `_id` field and generate your own event ID and store it as a string using something like the `uuid` lib.

events/app.py
```diff
@@ -2,7 +2,7 @@
 import datetime
 import json
 import pymongo
-from bson import json_util
+from bson import json_util, ObjectId
```
evankanderson: Where is `ObjectId` used?

mukobi: It's for serializing the `_id` field when searching for events by ID. MongoDB doesn't let you search for a field of type ObjectId by string, so you have to create an ObjectId yourself. See line 67.

evankanderson: D'oh... I totally missed this while scanning.

Thanks!

PR #51, opened 2019-08-02
(XS) Event picker for adding posts

cmei4444:
Better UI - Creates an easier way to assign an event to a post than a text field, displays event IDs and names in a dropdown menu.

### DIFF
--- a/pageserve/templates/index.html
+++ b/pageserve/templates/index.html
@@ -51,9 +51,13 @@ limitations under the License.
 
 <h2>Create New Post</h2>
 <form action="/v1/add_post" method="post" enctype="multipart/form-data">
-    <div> <!-- TODO(cmei4444): implement event picker -->
+    <div>
         <label for="event_id">Event ID:</label>
-        <input type="text" id="event_id" name="event_id">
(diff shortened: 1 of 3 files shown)


pageserve/templates/index.html
```diff
@@ -33,8 +33,8 @@ <h2>test user auth</h2>
 <h2>Past posts</h2>
 {% for post in posts %}
 <p>Posted by {{post.author_id}} at {{post.created_at}}</p>
-
-<p>For event {{ post.event_id }}</p>
+<!-- TODO(cmei4444) - link to event info -->
+<p>For event with ID {{ post.event_id }}</p>
```
evankanderson: Do you want to include this as a link instead?

cmei4444: I wanted to include an option to open up details for a selected event on a separate page so someone could check that they were posting to the right event. 
It's not a super important feature so I was going to leave it to another PR if I had time.

evankanderson: SGTM.

evankanderson:
LGTM

PR #52, opened 2019-08-02
(L) Delete posts endpoint

mukobi:
Add endpoint allowing for the deletion of posts.


posts/app.py
```diff
@@ -66,13 +66,26 @@ def get_all_posts():
...
 
-@app.route('/v1/<post_id>', methods=['GET'])
-def get_post_by_id(post_id):
-    """Get the post with the specified ID."""
-    # serialize otherwise nonserializable ObjectIDs
-    post_list = find_posts_in_db(
-        app.config["COLLECTION"], post_id=ObjectId(post_id))
-    return serialize_posts_to_json(post_list)
+@app.route('/v1/<post_id>', methods=['GET', 'DELETE'])
+def get_delete_post_by_id(post_id):
```
mayaspivak: It seems risky to have Get and Delete endpoints using the same method. It seems too easy to introduce a bug where you delete every post you try to get.  I would separate out into individual get and delete methods, at least for the outer methods. They could share an inner helper method if the code is mostly the same.

mukobi: Done https://github.com/knative-portability/large-events/pull/52/commits/951624c1dc9ebdd0faf98cc852d060df85b8abf3. I split the delete endpoint to `DELETE /v1/<post_id>/delete`.

posts/app.py
```diff
@@ -68,13 +68,29 @@ def get_all_posts():
...
-    """Get the post with the specified ID."""
-    # serialize otherwise nonserializable ObjectIDs
+    """Get the post with the specified ID.
+    """
     post_list = find_posts_in_db(
         app.config["COLLECTION"], post_id=ObjectId(post_id))
     return serialize_posts_to_json(post_list)
 
 
+@app.route('/v1/<post_id>/delete', methods=['DELETE'])
```
evankanderson: FWIW, you don't need to specify a different URL for deletion; the HTTP verb will indicate that (and it would be more aligned with REST to support DELETE at the same URL).

posts/app.py
```diff
@@ -83,6 +99,14 @@ def get_all_posts_for_event(event_id):
     return serialize_posts_to_json(post_list)
 
 
+def delete_post(post_id, author_id, collection):
+    """Deletes the post matching post_id and author_id if it exists."""
+    result = collection.delete_one(
+        {"_id": ObjectId(post_id), "author_id": author_id})
+    return (("Document deleted.", 204) if result.deleted_count
+            else ("Document not found.", 404))
```
evankanderson: Do you want to return a 403 if the `_id` matches but the `author_id` does not?

posts/test_post_deletion.py
```diff
@@ -0,0 +1,116 @@
...
+            post_id = post["_id"]
+            author_id = post["author_id"]
+            _, status_code = app.delete_post(
+                post_id, author_id, self.collection)
+            self.assertEqual(status_code, 204)
+        self.assertEqual(self.collection.count_documents({}), 0)
+
+    def test_not_existing_post_id(self):
+        """Can't find post by ID, don't delete."""
+        post_id = "C001""1C3D""C0FFEE""D0000000DE"   # invalid
```
evankanderson: Why multiple strings concatenated like this?

posts/app.py
```diff
@@ -68,13 +68,29 @@ def get_all_posts():
...
     return serialize_posts_to_json(post_list)
 
 
+@app.route('/v1/<post_id>/delete', methods=['DELETE'])
+def delete_post_by_id(post_id):
+    """Delete the post with the specified ID.
+
+    On deletion, only deletes posts matching post_id and with an author_id
+    matching the request parameter's author ID. Assumes that the passed in
+    author_id is authenticated and nonmalicies. I.e. this should only be
```
cmei4444: Small typo here - should it be "non-malicious"?

mukobi: Thanks, that's embarassing. Done https://github.com/knative-portability/large-events/pull/52/commits/7497b9b1b9ebe574403ac3a9f5ac0d0eb74aad27

posts/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/52/commits/7497b9b1b9ebe574403ac3a9f5ac0d0eb74aad27

posts/app.py
mukobi: I was thinking about that, but the query and response are just a lot simpler this way instead of handling 3 cases.

posts/test_post_deletion.py
mukobi: I was just trying to have fun and make the words more separated, though it works just like a single string. Changed https://github.com/knative-portability/large-events/pull/52/commits/7497b9b1b9ebe574403ac3a9f5ac0d0eb74aad27

PR #53, opened 2019-08-03
Created navbar and basic style improvements

cmei4444:
Created a navbar to move between pages. Also moved sign/in user buttons to navbar.

Changed font and image sizing app-wide.

### DIFF
--- a/pageserve/static/style.css
+++ b/pageserve/static/style.css
@@ -5,7 +5,7 @@
 }
 
 /* Style the links inside the navigation bar */
-.topnav a {
+.topnav .navtab {
   float: left;
   color: #f2f2f2;
   text-align: center;
@@ -15,18 +15,34 @@
 }
 
 /* Change the color of links on hover */
-.topnav a:hover {
+.topnav .navbtn:hover {
   background-color: #ddd;
   color: black;
 }
 
 /* Add a color to the active/current link */
-.topnav a.active {
+.topnav .navbtn.active {
   background-color: #4CAF50;
   color: white;
+}
+
+#user_nav {
+  float: right;
+}
+
+#sign_out_button {
+  border-left: 1px solid white;
+}
+
+#sign_in_button {
+  padding: 4% 8px;    /* TODO: vertically center sign-in button */
 }
 
 body {
   margin: 0px;
   font-family: Arial, Helvetica, sans-serif;
+}
+
+img {
+  width: 80%;
 }
--- a/pageserve/static/style.css
+++ b/pageserve/static/style.css
@@ -2,6 +2,9 @@
 .topnav {
   background-color: #333;
   overflow: hidden;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
 }
 
 /* Style the links inside the navigation bar */
@@ -26,21 +29,17 @@
   color: white;
 }
 
-#user_nav {
-  float: right;
-}
-
 #sign_out_button {
   border-left: 1px solid white;
 }
 
 #sign_in_button {
-  padding: 4% 8px;    /* TODO: vertically center sign-in button */
+  padding: 0px 8px;
 }
 
 body {
   margin: 0px;
-  font-family: Arial, Helvetica, sans-serif;
+  font-family: Helvetica, sans-serif, Arial;
 }
 
 img {
--- a/pageserve/static/style.css
+++ b/pageserve/static/style.css
@@ -1,0 +1,32 @@
+/* Add a black background color to the top navigation */
+.topnav {
+  background-color: #333;
+  overflow: hidden;
+}
+
+/* Style the links inside the navigation bar */
+.topnav a {
+  float: left;
+  color: #f2f2f2;
+  text-align: center;
+  padding: 14px 16px;
+  text-decoration: none;
+  font-size: 17px;
+}
+
+/* Change the color of links on hover */
+.topnav a:hover {
+  background-color: #ddd;
+  color: black;
+}
+
+/* Add a color to the active/current link */
+.topnav a.active {
+  background-color: #4CAF50;
+  color: white;
+}
+
+body {
+  margin: 0px;
+  font-family: Arial, Helvetica, sans-serif;
+}
--- a/pageserve/templates/base.html
+++ b/pageserve/templates/base.html
@@ -35,48 +35,43 @@ limitations under the License.
 
 <body>
     <div class="topnav">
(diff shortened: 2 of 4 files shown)


pageserve/static/style.css
```diff
@@ -0,0 +1,48 @@
+/* Add a black background color to the top navigation */
+.topnav {
```
cmei4444: So this navbar CSS styling was taken pretty directly from w3 schools' examples - is there any concern there?

mukobi: If it works, it works. Personally, I really passionately hate CSS `float`, where something like [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) is much cleaner.

pageserve/static/style.css
```diff
@@ -0,0 +1,48 @@
...
+#user_nav {
+  float: right;
+}
+
+#sign_out_button {
+  border-left: 1px solid white;
+}
+
+#sign_in_button {
+  padding: 4% 8px;    /* TODO: vertically center sign-in button */
```
mukobi: https://css-tricks.com/centering-css-complete-guide/

pageserve/static/style.css
```diff
@@ -0,0 +1,48 @@
...
+}
+
+body {
+  margin: 0px;
+  font-family: Arial, Helvetica, sans-serif;
```
mukobi: My vote here is `font-family: Helvetica, sans-serif, Arial;` or, better yet, importing a font from Google fonts.

pageserve/templates/base.html
```diff
@@ -32,44 +34,58 @@
 </head>
 
 <body>
+    <div class="topnav">
+        <a class="{% block home_tab %}{% endblock %} navtab navbtn" href='/v1/'>Home</a>
+        <a class="{% block events_tab %}{% endblock %} navtab navbtn" href='/v1/events'>Events</a>
```
mukobi: You might consider renaming `home_tab` and `events_tab` something like `home_tab_is_active` and `events_tab_is_active` or similar. On first reading, it was pretty unclear that you wanted these to toggle the `"active"` class.

pageserve/static/style.css
cmei4444: Changed to use flexbox in 91c9612

pageserve/static/style.css
cmei4444: Flexbox solved this in 91c9612

pageserve/static/style.css
cmei4444: Changed to first suggestion in 91c9612

pageserve/templates/base.html
cmei4444: Done in 91c9612

pageserve/static/style.css
evankanderson: If the actual text was copied (as opposed to simply used as an example), you should check the license. If you actually wrote it yourself based on w3 schools information, then it's your code to license as you see fit (and include a license header at the top).

Note that w3schools has an explicit license for fair use (with examples) but not an OSS copyright license:

https://www.w3schools.com/about/about_copyright.asp

pageserve/templates/base.html
```diff
@@ -32,44 +34,60 @@
...
+            {% block login_buttons %}
+            {% if not session['user'] %}
+            <div id="sign_in_button">
+                <div class="g-signin2" data-onsuccess="onSignIn" data-theme="dark"></div>
+                <script>
+                    function onSignIn(googleUser) {
+                        // Secure Google ID token to send to backend for verification.
+                        var id_token = googleUser.getAuthResponse().id_token;
+                        var xhr = new XMLHttpRequest();
+                        xhr.open('POST', "authenticate");
```
evankanderson: It looks like this is a relative URL based on the page -- if you don't make this a reference like `/authenticate`, make sure that the page can only be served from `/` and not from a path like `/post` (where the data might be sent somewhere else).

pageserve/templates/base.html
```diff
@@ -32,44 +34,60 @@
...
+        <div id="user_nav">
+            {% block login_buttons %}
+            {% if not session['user'] %}
+            <div id="sign_in_button">
+                <div class="g-signin2" data-onsuccess="onSignIn" data-theme="dark"></div>
+                <script>
+                    function onSignIn(googleUser) {
+                        // Secure Google ID token to send to backend for verification.
+                        var id_token = googleUser.getAuthResponse().id_token;
+                        var xhr = new XMLHttpRequest();
```
evankanderson: XHR is probably fine, but there's also the [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) if you want to use promises rather than callbacks.

pageserve/templates/base.html
```diff
@@ -32,44 +34,60 @@
...
+    <div style="padding: 0px 15px">
     <header>
         {% block header %}{% endblock header %}
     </header>
-    {% block login_buttons %}
-    {% if not session['user'] %}
-    <div id="sign_in_button">
-        <div class="g-signin2" data-onsuccess="onSignIn" data-theme="dark"></div>
-        <script>
-            function onSignIn(googleUser) {
```
evankanderson: It looks like something (probably the extra `div` tags) confused the github diff.

Not a big deal, but feel free to ignore comments above on re-indented but otherwise unchanged code.

pageserve/templates/events.html
```diff
@@ -35,8 +32,6 @@ <h1>{% block title %}Sub-Events{% endblock title %}</h1>
 <hr>
 {% endfor %}
 
-<a href='/v1/'>See all posts</a>
```
evankanderson: What happened to these links? I don't see them in `base.html` either.

pageserve/templates/index.html
```diff
@@ -15,20 +15,13 @@
...
 {% block content %}
-<h2>test user auth</h2>
-
-{% if auth %}
-<p>user has edit authorization</p>
-{% else %}
-<p>user does not have edit authorization</p>
-{% endif %}
-
-<a href='/v1/events'>See sub-events</a>
```
evankanderson: (ditto)

pageserve/templates/events.html
cmei4444: I actually moved them into the navbar, so there's no more need for links to other pages within the page body.

pageserve/templates/index.html
cmei4444: Moved to navbar with other link.

PR #56, opened 2019-08-05
(M) Pageserve delete post button

mukobi:
Add buttons to delete posts from pageserve front-end UI. For security, had to call a proxy route on pageserve which authenticates the user with get_user() and passes the user ID from that along to the posts deletion endpoint.


pageserve/templates/index.html
```diff
@@ -40,9 +60,13 @@ <h2>Past posts</h2>
 <p>{{post.text}}</p>
 {% endif %}
 {% if post.files %}
-  {% for file in post.files %}
-  <img src="{{ file }}">
-  {% endfor %}
+{% for file in post.files %}
```
evankanderson: Why flatten this indentation?

mukobi: Oh, that was unintentional. My autoformatter did that because all the HTML elements inside this jinja are siblings so this is technically correct. Should I unflatten?

cmei4444: I think it's much easier to read when unflattened - is there a way to disable the flattening?

pageserve/templates/index.html
```diff
@@ -33,9 +53,13 @@ <h2>Past posts</h2>
...
+<img src="{{ file }}">
+{% endfor %}
+{% endif %}
+
+{% if post.author_id == session["user_id"] %}
+<a href='#' onClick="{{'deletePost(\'' + post._id['$oid'] + '\');'}}">Delete</a>
```
cmei4444: This is a little hard to read, I think it's cleaner to just put `post._id['$oid']` in the braces:

```
onClick="deletePost('{{ post._id['$oid'] }}')"
```

pageserve/test_routes.py
```diff
@@ -32,6 +32,9 @@
 VALID_USER_AUTH_STATUS = 201
 
 EXAMPLE_USER = "app_user"
+EXAMPLE_USER_OBJECT = {
+    "user_id": "app_user",
+    "name": "app_user"}
```
cmei4444: Use `EXAMPLE_USER` here?

mukobi: See #58

pageserve/templates/index.html
mukobi: I just disabled my autoformatter for HTML files and reindented the file manually https://github.com/knative-portability/large-events/pull/56/commits/6568c90c905771f17b84959aa39f9a457eb4a955

pageserve/templates/index.html
mukobi: Thanks, much cleaner. Done https://github.com/knative-portability/large-events/pull/56/commits/e669f9dfbadb277761d5b45d24fe90e09c4acced

PR #60, opened 2019-08-06
(XL) Standardize quotation marks

mukobi:
Standardize quotation mark usage. Now all Python uses single quotes, except where a substring is to be in quotes in which case the whole string uses single quotes and the substring uses double quotes (e.g. `'She said, "Foobar."'`).


pageserve/templates/index.html
```diff
@@ -75,8 +75,7 @@ <h2>Create New Post</h2>
             <label for="event_id">Event ID:</label>
             <select id="event_id" name="event_id">
                 {% for event in events %}
-                    <!-- TODO (cmei4444) - update field names once other changes are pushed -->
-                    <option value="{{ event._id['$oid'] }}">{{ event._id["$oid"] }} - {{ event.name }}</option>
+                <option value="{{ event._id['$oid'] }}">{{ event._id['$oid'] }} - {{ event.name }}</option>
```
cmei4444: Sorry, just noticed your merge - Should there be double quotes on the second `"$oid"`?

mukobi: Done https://github.com/knative-portability/large-events/pull/60/commits/f8f41cd6804b79d0a3e1086aa6d1b277b1220f79

PR #61, opened 2019-08-06
(S) Make add post/event redirect to homepage

mukobi:
Currently, when you add a post or event using the forms on pageserve, you are sent to a page displaying the response from the posts/events service which is just a white page with the Mongo ID of the inserted data. This change still shows the response on failure, but if the post/event was successfully uploaded (status 201), it redirects you back to the index page.

### DIFF
--- a/pageserve/test_routes.py
+++ b/pageserve/test_routes.py
@@ -21,6 +21,7 @@ import ast
 import requests_mock
 from flask_testing import TestCase
 import flask
+from flask import url_for
 import app
 
 VALID_SESSION = {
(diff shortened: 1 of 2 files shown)


pageserve/test_routes.py
```diff
@@ -240,8 +240,7 @@ def test_add_valid_post(self, mock_requests):
 
         response = self.client.post('/v1/add_post', data=VALID_POST_FORM)
 
-        self.assertEqual(response.status_code, 201)
-        self.assertEqual(response.data.decode(), "Example success message")
+        self.assertEqual(response.status_code, 302)  # redirect to index
```
evankanderson: Check the location header?

mukobi: Done https://github.com/knative-portability/large-events/pull/61/commits/489d18ff8668d2620fa62e22460f856565a08dad

PR #62, opened 2019-08-06
(S) Add event check for authorization

mukobi:
Check for authorization before adding events. If authorized, normal behaviour. Otherwise, error message and status 403. Also through in a line that replaces the 'T' separator with a space in the `event_time` field from the add event form.


mukobi:
> Is this a change you want to make in the UI (pro: opportunity for better message / UX) or in the backing API (pro: consistent enforcement if there are several create/update paths through the UI)?

I chose to make this change in the UI because there is a `get_user()` function on pageserve that is tested and easily finds authorization whereas to implement authorization checking in the backing API would require copying over the functionality and testing of `get_user()` to the events service. Additionally, this is the same pattern used in adding and deleting posts where pageserve calls `get_user()` then sends the relevant information to the posts service. This also requires we mark those backing APIs (posts, events, users) as private and only accessible to pageserve when we deploy (e.g. "Require authentication" option on Cloud Run managed or limiting access to within the same cluster on Cloud Run for GKE).

pageserve/app.py
```diff
@@ -164,13 +164,21 @@ def add_event():
...
-    url = app.config['EVENTS_ENDPOINT'] + 'add'
-    form_data = dict(**request.form.to_dict(), author_id=get_user()['user_id'])
-    r = requests.post(url, data=form_data)
-    if r.status_code == 201:
-        # upload successful, redirect to index
-        return redirect(url_for("index"))
-    return r.content, r.status_code
+    try:
+        user = get_user()
+        if not user['is_organizer']:
```
cmei4444: I think this throws an unhandled TypeError if the user isn't logged in here.

pageserve/app.py
```diff
@@ -164,13 +164,21 @@ def add_event():
...
+        url = app.config['EVENTS_ENDPOINT'] + 'add'
+        form_data = dict(**request.form.to_dict(), author_id=user['user_id'])
+        # get rid of 'T' separator in event_time
+        form_data['event_time'] = form_data['event_time'].replace('T', ' ')
+        r = requests.post(url, data=form_data)
+        if r.status_code == 201:
+            # upload successful, redirect to index
+            return redirect(url_for('index'))
+        return r.content, r.status_code
+    except KeyError as error:
```
cmei4444: Should this be BadRequestKeyError instead? Or does it need to handle both?

pageserve/test_routes.py
```diff
@@ -32,9 +32,14 @@
 ERROR_BAD_TOKEN_STATUS = 400
 VALID_USER_AUTH_STATUS = 201
 
-EXAMPLE_USER_OBJECT = {
+AUTHORIZED_USER_OBJECT = {
     'user_id': 'app_user',
-    'name': 'app_user'}
+    'name': 'app_user',
+    'is_organizer': True}
```
cmei4444: I've noticed that some user objects use `'edit_access'` and others use `'is_organizer'` - we should standardize to one or the other.

cmei4444: Depending on how you did it for the users service, I think only some of my old pageserve code uses `'edit_access'` so it's probably best to replace that.

pageserve/app.py
mukobi: That throws a `KeyError` if the user isn't logged in (`user` is `None`) which is caught on line 180.

pageserve/app.py
mukobi: `BadRequestKeyError` won't be thrown since line 172 converts `request.form` to a dict which raises `KeyError`s on missing keys.

pageserve/test_routes.py
mukobi: I'll open a separate PR to convert everything to `'is_organizer'`.

pageserve/app.py
mukobi: Done https://github.com/knative-portability/large-events/pull/62/commits/d2d4d3477d13ac7079f75e694564a02f23be8c03

pageserve/test_routes.py
mukobi: #68

pageserve/app.py
cmei4444: That works then!

PR #63, opened 2019-08-06
(S) Pageserve handle downed users service

mukobi:
Previously, pageserve just cried and died when the users service was down. Hopefully this isn't really needed (makes local development easier though), but now pageserve is a strong, independent service that won't break down the millisecond it can't find its sibling.


mukobi:
> LGTM - if the users service was down would the user be able to tell? Although at this point I think it's fine if it just looks like they aren't logged in.

No, they'd just stay logged out. There's a tradeoff where this way it might be a little harder to tell if the users service is down but if it is you can still use and test the rest of the app, which I thought was better than crashing the server on a requests `ConnectionError`.

PR #68, opened 2019-08-07
(S) Change edit access to is organizer

mukobi:
Standardize use of `'is_organizer'` rather than mix of that and `'edit_access'` for authorization status.

### DIFF
--- a/events/app.py
+++ b/events/app.py
@@ -67,8 +67,6 @@ def add_event():
             'author': request.form['author_id'],
             'event_time': request.form['event_time']
         }
-        # TODO(cmei4444): Athenticate user and verify that user has event
-        # editing access
         current_time = datetime.datetime.utcnow().isoformat(sep=' ',
                                                             timespec='seconds')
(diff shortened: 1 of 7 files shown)


events/app.py
```diff
@@ -67,8 +67,6 @@ def add_event():
             'author': request.form['author_id'],
             'event_time': request.form['event_time']
         }
-        # TODO(cmei4444): Athenticate user and verify that user has event
```
evankanderson: Presumably because we changed the pattern of where authorization is requested in #62?

mukobi: Correct, see https://github.com/knative-portability/large-events/pull/68/commits/c679c15d4dad5b8f38ee1e9bb7cb77858e5edf5e

mukobi:
> Have you looked at what happens if the change to the users service is rolled out, but the change to pageserve is not (or vice versa)?

That should break it since we'd have mismatching APIs, but this changes both at the same time so it should be good.

PR #69, opened 2019-08-07
(S) Better ui

mukobi:
Make the UI a little better.
- Center images
- Limit width of images
- Enable nav-bar wrapping on small mobile devices
- Make search bar width a bit more dynamic

https://screenshot.googleplex.com/C1C6NkoCSvn
https://screenshot.googleplex.com/OreGCKDBZxQ

### DIFF
--- a/pageserve/static/style.css
+++ b/pageserve/static/style.css
@@ -3,6 +3,7 @@
   background-color: #333;
   overflow: hidden;
   display: flex;
+  flex-wrap: wrap;
   justify-content: space-between;
   align-items: center;
 }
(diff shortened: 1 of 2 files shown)


pageserve/static/style.css
```diff
@@ -3,6 +3,7 @@
   background-color: #333;
   overflow: hidden;
   display: flex;
+  flex-wrap: wrap;
```
cmei4444: How will this look if the user is signed in? (since there will be an extra box)

mukobi: The second line is just a bit longer. flex-wrap automatically wrap flexbox elements to the next row if they don't fit on the current one.

PR #70, opened 2019-08-07
(S) Add post check authentication

mukobi:
Make add post pageserve route return error if user is not logged in (previously crashed with a TypeError). Also disables the add post button if the user is not signed in.

### DIFF
--- a/pageserve/test_routes.py
+++ b/pageserve/test_routes.py
@@ -350,6 +350,13 @@ class TestAddPostRoute(unittest.TestCase):
         self.assertEqual(response.status_code, 400)
         self.assertEqual(response.data.decode(), 'Example error message')
 
+    def test_not_logged_in(self):
+        """Not logged in, don't add."""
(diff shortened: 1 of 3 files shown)


pageserve/test_routes.py
```diff
@@ -350,6 +350,13 @@ def test_add_invalid_post(self, mock_requests):
         self.assertEqual(response.status_code, 400)
         self.assertEqual(response.data.decode(), 'Example error message')
 
+    def test_not_logged_in(self):
+        """Not logged in, don't add."""
+        response = self.client.post('/v1/add_post', data=INVALID_POST_FORM)
```
cmei4444: I think it would be better to use `VALID_POST_FORM` here so you're only testing a missing user.

evankanderson: +1

mukobi: Thanks for the catch, done https://github.com/knative-portability/large-events/pull/70/commits/14517d48654c7704dd2675d199c0942eef42f66d

PR #71, opened 2019-08-07
(S) Text search for searching events by name

cmei4444:
Using the text search functionality from MongoDB, searching events by name is less restrictive - capitalization is ignored by default.

Another built in feature of text search is that it searches on word stems (searching 'events' will also return the name 'event' if it exists) - could be good or bad, but I think it's convenient if a person doesn't know exact names.
It also ignores stop words (e.g 'the', 'and')

### DIFF
--- a/events/test_routes.py
+++ b/events/test_routes.py
@@ -174,6 +174,20 @@ class TestSearchEventsRoute(unittest.TestCase):
         self.assertEqual(len(data['events']), 1)
         self.assertEqual(data['num_events'], 1)
 
+    def test_search_existing_event_uppercase(self):
+        """Search for an event that exists with different capitalization.
+
+        The event should be found because search is case-insensitive.
+        """
+        response = self.client.get(
+            '/v1/search?name=' + VALID_EVENT_NAME.upper())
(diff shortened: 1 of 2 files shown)


events/test_routes.py
```diff
@@ -174,6 +176,22 @@ def test_search_existing_event(self):
         self.assertEqual(len(data['events']), 1)
         self.assertEqual(data['num_events'], 1)
 
+    @patch('app.text_search_event_name', MagicMock(return_value=[VALID_DB_EVENT]))
+    def test_search_existing_event_uppercase(self):
+        """Search for an event that exists with different capitalization.
```
evankanderson: If you've switched to a mock, you're no longer testing the behavior of Mongo; instead, you might want to [validate the arguments to `text_search_event_name`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args):

```python
aself.assertEqual(VALID_EVENT_NAME.upper(), app.text_search_event_name.call_args[0][1])
```

cmei4444: Done in c6279ae

PR #72, opened 2019-08-07
(M) Created links to events in posts

cmei4444:
Each post now has a link to the event associated with it, bringing up all of the info about the event.

Created endpoint to handle querying and tests for the endpoint. 
Also made some small fixes to quotes/doctrings in TestSearchEventsRoute test class.

### DIFF
--- a/pageserve/test_routes.py
+++ b/pageserve/test_routes.py
@@ -60,6 +60,7 @@ EXAMPLE_POSTS = [
      'event_id': 'valid_post_id',
      'text': 'example post 2.'}]
 EXAMPLE_EVENT_ID = '123456789123456789123456'
+DIFFERENT_EVENT_ID = '987654321987654321654321'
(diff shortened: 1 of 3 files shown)


pageserve/test_routes.py
```diff
@@ -59,14 +59,15 @@
...
 EXAMPLE_EVENTS = [
-    {'_id': {'$oid': '123abc'},
+    {'_id': EXAMPLE_EVENT_ID,
      'name': 'valid_event',
      'description': 'this event is valid',
      'event_time': 'soon',
      'author': 'app_user',
      'created_at': 'in the past'},
-    {'_id': {'$oid': '456def'},
+    {'_id': '987654321987654321654321',
```
mayaspivak: nit: define this as a constant too.

cmei4444: Done in 4831809

PR #74, opened 2019-08-07
(S) Fix no image posting

mukobi:
Fixed posts trying to make an image out of nothing when you don't upload an image from the form, and in the process enabled multiple image uploading.

### DIFF
--- a/posts/app.py
+++ b/posts/app.py
@@ -204,7 +204,8 @@ def upload_new_post_to_db(post, collection):
     # post is valid, add on timestamp, upload files, insert into db
     post['created_at'] = generate_timestamp()
     post['files'] = [
-        upload_file_to_cloud(file) for file in post['files']]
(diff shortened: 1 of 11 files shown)


posts/app.py
```diff
@@ -204,7 +204,8 @@ def upload_new_post_to_db(post, collection):
     # post is valid, add on timestamp, upload files, insert into db
     post['created_at'] = generate_timestamp()
     post['files'] = [
-        upload_file_to_cloud(file) for file in post['files']]
+        upload_file_to_cloud(file) for file in post['files']
+        if file.filename != '']
```
cmei4444: Would it be better to handle this check in pageserve?

mukobi: Yeah, sorry, fixed in https://github.com/knative-portability/large-events/pull/74/commits/71bd359d2ff3be1c085d33837db1225be4d4245a

PR #76, opened 2019-08-08
(S) Front-end changes

cmei4444:
Most changes were made before demo - just some minor front-end improvements.

- Made navbar stay visible at the top of the screen even when scrolling
- Moved post submit form to top of page for better flow, especially since posts are in reverse order

### DIFF
--- a/pageserve/static/style.css
+++ b/pageserve/static/style.css
@@ -1,10 +1,13 @@
 /* Add a black background color to the top navigation */
 .topnav {
+  width: 100%;
   background-color: #333;
   overflow: hidden;
   display: flex;
   justify-content: space-between;
   align-items: center;
+  position: fixed;
+  top: 0;
 }
 
 /* Style the links inside the navigation bar */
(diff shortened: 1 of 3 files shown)


pageserve/static/style.css
```diff
@@ -139,7 +142,7 @@ textarea {
 }
 
 .page_content {
-  margin: 0 15%;
+  margin: 75px 15% 30px;
```
mukobi: Did you test this on mobile too? The top bar might wrap over and overlap content if this isn't big enough. You can use the device emulators in Chrome dev tools (F12 then Ctrl+Shift+M) to test.

mukobi: If it's not enough an easy fix might be to find the width where the header wraps then do
```css
@media screen and (max-width: width-in-px) {
  .page_content {
    margin-top: 150px;
  }
}
```

cmei4444: Done e99c822, looks like https://screenshot.googleplex.com/H8LQfUb5ygn on small screens.

PR #79, opened 2019-08-08
Added UI for filtering posts by event

cmei4444:
Created links in each event to all the posts created for that event. Also added some additional logic in the template to tell the user what subset of posts they're looking at.

Also created a dropdown for filtering by event on the home page, but was running into some formatting issues:
Any idea how to right-align this? https://screenshot.googleplex.com/idNBs4wv8g5
Using flexbox to right align causes the content to overlap the dropdown button due to its absolute positioning: https://screenshot.googleplex.com/9LAAfCiBVuU

### DIFF
--- a/pageserve/app.py
+++ b/pageserve/app.py
@@ -99,6 +99,7 @@ def get_posts_for_event(event_id):
             posts=parse_posts(response.json()),
             auth=is_organizer(get_user()),
             events=get_events(),
+            sub_event=event_id,
             app_config=app.config
         )
     else:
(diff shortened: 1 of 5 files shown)


cmei4444:
> To do the right align without overlapping the dropdown button, would this work?
> 
> ```css
> .dropdown-content {
>   position: absolute;
>   top: 32px;  /* or height of dropdown button */
> }
> ```

This does work when the screen is wider, but on smaller screens the text in the button wraps so the button height changes. 
I could just check for the screen width where that happens and change the position there.

PR #80, opened 2019-08-12
Add clear filter button

mukobi:
Add button to clear filter after filtering by event
https://screenshot.googleplex.com/1VYM8XfJhmF

Last PR I promise!

### DIFF
--- a/pageserve/templates/index.html
+++ b/pageserve/templates/index.html
@@ -75,9 +75,8 @@ limitations under the License.
     <hr class="form_divider">
 
     {% if sub_event %}
-    <h2>Past Posts for Event {{ sub_event }}</h2>
     <div class="filter_bar">
-        <div></div>
+        <h2>Past Posts for Event {{ sub_event }}</h2>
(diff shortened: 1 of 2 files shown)


pageserve/templates/index.html
```diff
@@ -76,11 +76,17 @@ <h2>Create New Post</h2>
 
     {% if sub_event %}
     <h2>Past Posts for Event {{ sub_event }}</h2>
+    <div class="filter_bar">
+        <div></div>
```
cmei4444: Might be cleaner if you put the "Past Posts for Event [event]" in the filter bar too, like on the non-subevent case below.

The downside is that the formatting might be strange on smaller screens.

mukobi: Thanks, done https://github.com/knative-portability/large-events/pull/80/commits/cca265355d44bae15f6b54e5e2bdcb88a5facfd9
https://screenshot.googleplex.com/wF33frHDXJp

